PR #117, opened 2020-08-18
Add auto-deploy to staging job

cicdw:
PR Checklist:

## Describe this PR
This PR adds a new CircleCI workflow that builds and deploys the UI to our staging environment with every merge to master.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -76,6 +76,30 @@ jobs:
           at: .
       - deploy_env:
           BUCKET_NAME: ${GCS_BUCKET__DEV}
+  build-staging:
+    resource_class: large
+    docker:
+      - image: node:14.7
+    steps:
+      - build:
+          VUE_APP_AUTH0_DOMAIN: ${VUE_APP_AUTH0_DOMAIN__STAGING}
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -76,6 +76,30 @@ jobs:
...
+    docker:
+      - image: node:14.7
+    steps:
+      - build:
+          VUE_APP_AUTH0_DOMAIN: ${VUE_APP_AUTH0_DOMAIN__STAGING}
+          VUE_APP_AUTH0_PUBLIC_CLIENT_ID: ${VUE_APP_AUTH0_CLIENT_ID__STAGING}
+          VUE_APP_STRIPE_PUBLIC_TOKEN: ${VUE_APP_STRIPE_PK__STAGING}
+          VUE_APP_CLOUD_URL: ${VUE_APP_CLOUD_URL__STAGING}
+          VUE_APP_BASE_URL: ${VUE_APP_BASE_URL__STAGING}
+      - persist_to_workspace:
```
cicdw, line 90: I'll admit I'm not 100% sure if this persist step will conflict with the one in dev for any reason

znicholasbrown: I don't think it should - shouldn't that persist only between jobs in the workflow and not between workflows?

cicdw: Yea that would be my expectation

PR #127, opened 2020-08-21
Unit tests store api

### LINKED ISSUE #144: Write unit test for the API Vuex Store
**Why**
Increase our test coverage and give us increased confidence in the api store

**What**
* [ ] Test for state
* [ ] Test getters
* [ ] Test mutations
* [ ]  Test actions
* [ ] Add type checks to the store for backend choices 

**Notes**
Sub issue of #101

zhen0:
PR Checklist:

## Describe this PR
- Adds unit tests for the api Vuex store
- Closes #144


tests/unit/store/api.spec.js
```diff
@@ -0,0 +1,635 @@
...
+        // that we don't want to
+        // test here
+        api.mutations['tenant/setTenant'] = jest.fn()
+        api.mutations['tenant/setDefaultTenant'] = jest.fn()
+        api.mutations['tenant/unsetTenants'] = jest.fn()
+        api.mutations['tenant/unsetTenant'] = jest.fn()
+        api.actions['tenant/getTenants'] = jest.fn()
+        api.actions['auth0/authenticate'] = jest.fn()
+        api.actions['auth0/authorize'] = jest.fn()
+        api.actions['user/getUser'] = jest.fn()
```
znicholasbrown, line 564: 👍

tests/unit/store/api.spec.js
```diff
@@ -0,0 +1,635 @@
...
+      it('should set the connection message', async () => {
+        await store.dispatch('monitorConnection')
+        expect(store.getters.connectionMessage).toBe('Connected')
+      })
+      it('should set retries back to 0', async () => {
+        await store.dispatch('monitorConnection')
+        expect(store.getters.retries).toBe(0)
+      })
+    })
+    //Mock won't work as error policy set to 'none'
```
znicholasbrown, line 509: If we can't test this part of the store, should we change the error policy on the `fallbackApolloClient` to let us catch issues? Doesn't necessarily need to be in this PR but seems like something we should be able to test.

tests/unit/store/api.spec.js
```diff
@@ -0,0 +1,635 @@
...
+    })
+    it('should return the active url', () => {
+      expect(store.getters.url).toBe(undefined)
+    })
+    it('should return the version', () => {
+      expect(store.getters.version).toBe(null)
+    })
+  })
+
+  describe('where relevant getters check local storage before environment variables', () => {
```
znicholasbrown, line 189: I don't think this is a true statement - none of the getters for the API store check `localStorage` or environment variables; only the initial state is set based on those. As such, I think just the above `State`-testing describe block will cover this, meaning we can combine all 3 getters describe blocks into 1 general use pattern.

tests/unit/store/api.spec.js
zhen0: Actually, I updated the mock and can now test this.  We also added a catch for if no data returned.  Tested locally and can't see any error.  (Not actually sure where we use that connectionMessage?)

zhen0:
Ready for another review @znicholasbrown

PR #163, opened 2020-08-28
State manipulation improvements

### LINKED ISSUE #141: ENH: Ability to place runs into *any* state
**What and Why**
I find myself experimenting a lot in the UI to test features, etc. in doing so, I sometimes want to manually place runs into various states (e.g., `Scheduled` or `Pending`) but the drop down lists in the UI don't always allow me to do so.  In that case I find myself having to go back and forth from the GraphQL API - it would be nice if the UI didn't place any restrictions on the types of states I could set.

cicdw:
PR Checklist:

## Describe this PR
This PR improves the UX of state setting in the following ways:
- enhances the organization of what states you can "mark" task and flow runs as - the lists are based on most used + those that are not internal application-logic-specific.  While it doesn't expose every possible state, I argue it closes #141 
- Allows users to restart a Flow Run from "Failed" anytime the Flow run itself is in a Failed state (this is useful in situations where the original deployment of the flow run failed)
- Fixes a bug in the state messages + log generation when using Restart
- removes stale references to result handlers; while I do think this message was originally useful, I think there are many cases in the newer versions of Prefect where either a default result was pre-configured, individual tasks (instead of the Flow) have results configured, or the tasks don't exchange data at all.  Because of this, I think that a simpler dialogue is better for now.


src/pages/FlowRun/Restart-Dialog.vue
```diff
@@ -151,40 +157,19 @@ export default {
       Restart from failed?
     </v-card-title>
 
-    <v-card-text v-if="hasResultHandler">
+    <v-card-text>
       Click on confirm to restart
       <span class="font-weight-bold">{{ flowRun.name }}</span> from its failed
```
zhen0, line 162:
```suggestion
      <span class="font-weight-bold">{{ flowRun.name }}</span> 
```

src/pages/FlowRun/Restart-Dialog.vue
```diff
@@ -151,40 +157,19 @@ export default {
...
       task run(s).
```
zhen0, line 163:
```suggestion

```

src/pages/FlowRun/Restart-Dialog.vue
zhen0: Since we can now restart without a failed task run, we should maybe take this bit out.

src/pages/FlowRun/Restart-Dialog.vue
```diff
@@ -196,7 +181,7 @@ export default {
         <span v-if="role === 'READ_ONLY_USER'">
           Read-only users cannot restart flow runs
         </span>
-        <span v-else-if="!failedTaskRuns">
+        <span v-else-if="!isFailedRun">
           You can only restart a flow run with failed tasks.
```
zhen0, line 185:
```suggestion
          You can only restart a failed flow run.
```

cicdw:
@zhen0 good catches - I committed them myself so that I could lint them

PR #176, opened 2020-08-31
Order params

### LINKED ISSUE #126: Ensure parameter ordering in the UI
## Current behavior
The order that flow parameters appear in the Ui seems a) to be arbitrary and b) may change on flow registration.  This has a couple of drawbacks:
- Sometimes there is a logical ordering to the parameters- e.g. it makes sense that a "start" parameter should come before an "end" parameter.
- Flow parameter order jumping around can mean it's easy to mistake one parameter for another.

## Proposed behavior
- Allow the user to define the order of parameters- maybe the easiest thing here is to have them show up here in the same order they are defined in the flow. 
- If the above is not possible for any reason then ensure that the ordering is well defined and does not change on flow registration.  One option here is to have the order be alphabetical.

## Example
I have three parameters, `start`, `end` and `environment`.  If this proposal is implemented I can ensure that the ui will always present them in this order, which makes it easier for someone scheduling a flow to fill in the parameters correctly.

zhen0:
PR Checklist:

## Describe this PR
- Adds sort for default parameters 
- Uses default parameters to show parameters in the flow details tile (as these are the ones that will be used if a flow is run)
- Adds an info box to the parameters section of the flow details tile with a link to the docs and the flow settings page
- Closes #126


znicholasbrown:
This looks great @zhen0 - just a minor bug:
Screen Shot 2020-08-31 at 5 18 45 PM

The menu doesn't close when navigating to the in-app links

zhen0:
Thanks @znicholasbrown - should be ready for another review.

Issue #175, opened 2020-08-31
Overenthusiastic scrollbars in both cloud and server UI


jnoynaert:
**Bug Description**

The most recent version of the UI is displaying extra scrollbars for everything that uses Vue list item groups. 
- This occurs when the UI is rendered in Chrome, Firefox, or Edge.
- This appears to affect Windows only.
- Tested on both cloud and server (from latest core release).

Adding the following CSS rule to override the default styling (`overflow:scroll`) fixes it for me:

```css
.v-list-item__content {
 overflow:auto
}
```

**Steps To Reproduce**

Using Firefox or Edge on a Windows machine or VM, visit the overview page for a specific flow (this issue appears to affect most pages of the UI, but is easily obvious on this page).

**Browsers Tested:**
* [X] Chrome (latest)
* [X] Firefox (80.0)
* [ ] Safari
* [X] Edge (85.0.564.41)
* [ ] IE

----------
**Notes**

![image](https://user-images.githubusercontent.com/5920864/91746631-69ca7680-ebac-11ea-9b2e-bd573baf8a49.png)

znicholasbrown:
Thanks for reporting this @jnoynaert - I'm taking a look at this issue today once I can get a Windows machine set up.

PR #177, opened 2020-08-31
Scrollbars update

znicholasbrown:
PR Checklist:

## Describe this PR
Overrides (non-Mac) platform and browser-specific scrollbars styling for Firefox and Webkit-based browsers - scrollbars should no longer be visible at all times for those browsers and should more closely match original design specs. The exception to this is Firefox, which would require a heavier JavaScript-based solution so went with a "close-enough" CSS fix instead.

Resolves: #175

cc @jnoynaert

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -4,11 +4,11 @@
 
 ### Features and Improvements
 
-- None
+- Update scrollbars with platform-specific styling [#177](https://github.com/PrefectHQ/ui/pull/177)
 
 ### Bugfixes
 
-- None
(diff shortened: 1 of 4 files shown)


znicholasbrown:
**Edge**

Before:
edge-before

After:
edge-after

**Firefox**

Before:
ff-before

After:
ff-after

**Chrome**

Before:
chrome-before

After:
chrome-after

Note: only modern versions of the above browsers were tested, since that's what we support.

zhen0:
QA'd this using Lambdatest (https://www.lambdatest.com/) and it looks good to me - @znicholasbrown 
Need to update the changelog though.

PR #179, opened 2020-08-31
Display # of mapped children (when available)

### LINKED ISSUE #173: Show the total number of mapped tasks that there will be
**Why**
This is to show the number of mapped tasks that there will be. Dependent on [PR#3235](https://github.com/PrefectHQ/prefect/pull/3235), created because of point 1 in [prefect/3233](https://github.com/PrefectHQ/prefect/issues/3233). 

See note below from 3233. 

"There's no way in the UI to see how many mapped tasks there will be. Mapped tasks appear as they discovered, even though (in principle) at the point that the parent state transitions to Mapped, it's possible to know the total number of mapped children that will appear."

**Notes**

cicdw:
PR Checklist:

## Describe this PR
This PR displays the _expected_ number of mapped tasks to be generated by a mapped pipeline.  Note:
- I use the term "expected" because if something goes wrong, the number could be smaller
- this data is only available on Core versions 0.13.5+ (note this has not been released yet)

I additionally chose to remove `Class` as a displayed field.  It's Python specific and more often than not too generic to be useful.

Closes #173

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -4,7 +4,7 @@
 
 ### Features and Improvements
 
-- None
+- Display number of expected mapped runs when the data exists - [#173](https://github.com/PrefectHQ/ui/issues/173)
 
 ### Bugfixes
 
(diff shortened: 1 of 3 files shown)


znicholasbrown:
Looks good @cicdw - can we add to the tooltip a small explanation of the situations in which the number of mapped runs ends up being less than the displayed expected number?

PR #187, opened 2020-09-01
Fix mapped children restart from failure bug

### LINKED ISSUE #182: Bug with checkpointing and mapped tasks
## Description

Currently restarting a failed flow that has failed on some mapped tasks seems wrong in the following ways:

1. Previously successful mapped childs will re-run
2. `run_count` doesn't seem to change between restarts

## Expected Behavior

1. A previously successful mapped task should not be rerun by checkpointing if it has already succeeded
2. the `run_count` should up between runs.

## Reproduction

Here is a minimal flow that can reproduce this is below. Start a cloud run, it will fail on the first mapped task, but the other two will succeed. Use the 'restart' button in the top right of the UI to restart this flow-run. Note that previously successful mapped runs run again. Note that the first instance fails again, and (in the logs) `run_count` is still `1`, even though the UI clearly shows multiple runs.

```python

import prefect

@prefect.task
def task(x):
    prefect.context.get('logger').info(f'x = {x} and run_count = {prefect.context.get("task_run_count")}')
    if x == 0 and prefect.context.get('task_run_count') == 1:
        raise prefect.engine.signals.FAIL('fail on first attempt')
    else:
        return x + 1

@prefect.task
def summy(xs):
    return sum(xs)

with prefect.Flow('test-restarts') as flow:
    list = prefect.tasks.core.constants.Constant([0,1,2])

    mapped = task.map(list)

    summy(mapped)

flow.register(project_name='mattk')
```

## Environment

```
root@docker-desktop:/workspaces/prefect# prefect diagnostics
{
  "config_overrides": {},
  "env_vars": [],
  "system_information": {
    "platform": "Linux-4.19.104-microsoft-standard-x86_64-with-debian-10.5",
    "prefect_version": "0.13.2",
    "python_version": "3.7.9"
  }
}
```

cicdw:
PR Checklist:

## Describe this PR
Closes #182 

This PR adds some nuance to the restart from failed button: in particular, because our utility call that traverses the Flow is based on _task_ IDs, we were previously incorrectly resetting the state of mapped pipelines that had partially succeeded.  This PR fixes that by handling mapped pipelines in a special way: if there is a map_index present on the run, we explicitly check if it was a failure, otherwise we don't need to reset its state.

### DIFF
--- a/src/pages/FlowRun/Restart-Dialog.vue
+++ b/src/pages/FlowRun/Restart-Dialog.vue
@@ -37,13 +37,27 @@ export default {
 
         this.writeLogs()
 
-        const taskRunStates = this.utilityDownstreamTasks.map(task => {
-          return {
-            version: task.task.task_runs[0].version,
-            task_run_id: task.task.task_runs[0].id,
-            state: { type: 'Pending', message: this.restartMessage }
-          }
-        })
+        // we want to avoid resetting the parent Mapped state of a mapped pipeline,
+        // as that would cause *all* children to rerun, regardless of whether they failed.
+        // So, first we collect all candidate run states and then filter:
+        // if map_index is null, we can proceed normally - if not null, we explicitly
+        // check whether it's one of the failed states
+        let failedRunIds = this.failedTaskRuns.map(run => run.id)
+
+        const taskRunStates = this.utilityDownstreamTasks
+          .map(task =>
+            task.task.task_runs.map(run => {
+              if (!run.map_index || failedRunIds.includes(run.id)) {
+                return {
+                  version: run.version,
+                  task_run_id: run.id,
+                  state: { type: 'Pending', message: this.restartMessage }
+                }
+              }
+            })
+          )
+          .flat()
+          .filter(x => x)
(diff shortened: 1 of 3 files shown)


src/pages/FlowRun/Restart-Dialog.vue
```diff
@@ -37,13 +37,27 @@ export default {
...
+                return {
+                  version: run.version,
+                  task_run_id: run.id,
+                  state: { type: 'Pending', message: this.restartMessage }
+                }
+              }
+            })
+          )
+          .flat()
+          .filter(x => x)
```
znicholasbrown, line 60:
```suggestion
          // the above flat map doesn't always return objects
          // but can return something like [undefined, {...}, undefined]
          // so we filter falsey values here
          .filter(x => !!x)
```
Let's add a double bang and a comment here to make it clear that we're filtering any falsey values (this wasn't immediately obvious).

cicdw: 👍  makes sense

PR #188, opened 2020-09-01
Filtering scheduled flows from the heartbeattimeline

kingsleyb:
PR Checklist:

## Describe this PR

The dual timestamps on the activity timeline for scheduled flow runs is causing some confusion on the meaning of the two displayed times. Given that the scheduled flow runs are shown both in the dashboard overview and also the flow run over view in a separate window, I want to suggest removing the scheduled flow runs from the activity timeline altogether. 

This PR shows how this would look. 

I'll wait on feedback on the suggested change prior to adding to changelog.


kingsleyb:
Made some changes based on the feedback provided on where to filter out the Scheduled states.

src/graphql/Dashboard/heartbeat.gql
```diff
@@ -1,10 +1,10 @@
-query Heartbeat($projectId: uuid, $timestamp: timestamptz, $state: String) {
+query Heartbeat($projectId: uuid, $timestamp: timestamptz, $state: String, $filterOutState: String) {
   flow_run(
     limit: 10
     where: {
       flow: { project_id: { _eq: $projectId } }
       state_timestamp: { _gte: $timestamp }
-      state: { _eq: $state }
+      state: { _eq: $state, _neq: $filterOutState }
```
jlowin, line 7: This feels a little convoluted and will be confusing when separated from the motivation in this PR, because if someone weren't aware of how Hasura treats `null` variables, this looks like we're saying "give me flow runs that match a state but not a different state", which doesn't make much sense.

However, I recognize that it fits our usage. My suggestion would be to make it a `_nin` clause instead of a `_neq` clause, to make it easier to filter out potentially other states going forward. 
```suggestion
      state: { _eq: $state, _nin: $filterOutStates }
```

src/graphql/Dashboard/heartbeat.gql
```diff
@@ -1,10 +1,10 @@
-query Heartbeat($projectId: uuid, $timestamp: timestamptz, $state: String) {
+query Heartbeat($projectId: uuid, $timestamp: timestamptz, $state: String, $filterOutState: String) {
```
jlowin, line 1:
```suggestion
query Heartbeat($projectId: uuid, $timestamp: timestamptz, $state: String, $filterOutStates: [String!]) {
```

jlowin (changes requested):

src/graphql/Dashboard/heartbeat.gql
kingsleyb: @jlowin this makes sense but should $state also be made a list whilst i'm making this change?

jlowin: I'm more hesitant to touch things that are already in use because it may require more updates elsewhere

src/graphql/Dashboard/heartbeat.gql
kingsleyb: made suggested changes.

jlowin (commented):
👍

jlowin (dismissed):

PR #200, opened 2020-09-04
Show the token name on the agent card instead of the token id

kingsleyb:
PR Checklist:

## Describe this PR
Showing the token name in the agent card instead of the token id will make it easier to identify which token the agent is using.


src/components/Agents/AgentCard.vue
```diff
@@ -270,41 +270,41 @@ export default {
...
                     x-small
                     class="mb-2px mr-2"
                     tabindex="0"
                     color="white"
                   >
-                    {{ copiedText[agent.token_id] ? 'check' : 'file_copy' }}
+                    {{ copiedText[agent.token_name] ? 'check' : 'file_copy' }}
                   </v-icon>
                   {{
-                    agent.token_id
```
jlowin, line 305: Since token names are not unique, it may still be helpful to allow copying the ID rather than the name, for any queries.

cicdw: As of right now, the new Agent persistence PRs do not expose token id as a queryable field on agents

cicdw: @znicholasbrown FWIW @joshmeek and I agreed earlier to persist both token id and name in the DB in case we want to revisit its use

znicholasbrown: Ok scratch that then @kingsleyb - no need to remove it from the query.

znicholasbrown (approved):

PR #201, opened 2020-09-04
Performance work

znicholasbrown:
PR Checklist:

## Describe this PR
Makes some changes to the barchart visualization to take browser visibility into account when performing repaints of the canvas and svg elements. 

Removes `task_run` joins on the upcoming runs queries for both the dashboard and flow pages. 

Creates the scaffolding for a better service worker infrastructure for heavier data visualizations and eventual token refreshes. 

Adds extended slots to the TileLayout (primarily for testing, we won't use these yet)

Simplifies our duration calculations by removing the reliance on `moment` and using the JavaScript native Date object instead; this should be a nice performance improvement itself since instantiating a new `moment` object for each of our timestamps is pretty heavy. Applies to the upcoming/late runs tiles, duration spans, timeline mixins, and elsewhere. 

Exports the cache from the `vue-apollo` plugin (also renames it to avoid confusion).


zhen0 (approved):
LGTM @znicholasbrown!

PR #212, opened 2020-09-08
Performance work part 2

znicholasbrown:
PR Checklist:

## Describe this PR
Should drastically improve app performance when left in background tabs and prevent large memory leaks when navigating between pages.

Edit: this should also cut the initial memory load by ~50%.


src/main.js
```diff
@@ -146,15 +147,63 @@ Vue.filter('roundThousands', roundThousands)
 
 Vue.component('height-transition', TransitionHeight)
 
+// This is a global mixin used to clean up any
+// references a component may have after it's destroyed.
+// It's a pretty heavy-handed approach to what we're trying
+// to accomplish but Vue doesn't seem to be removing all references
+// itself, which is leading to large memory leaks on data-heavy pages
+// as Apollo polls and when navigating between pages.
+// This could be more elegant and may be fixed by upgrading to Vue 3.
```
znicholasbrown, line 156: Note for the reviewer: this will knowingly flood the console with errors when certain components are destroyed. This happens because Vue really doesn't like when you modify some of these reactive props and may even have a `$nextTick` queued after the component is destroyed (or something). The errors are innocuous for now but once this is out and we can test it in lower environments I'll take another pass at cleaning those up and/or narrowing this list.

zhen0 (approved):
Thanks @znicholasbrown - LGTM

PR #216, opened 2020-09-10
Barchart bugfix VU

znicholasbrown:
PR Checklist:

## Describe this PR
Adds an optional `visible` prop to the `BarChart` component, which prevents resize and update handlers from firing when set to false. FIxes instances where timeline tiles end up in states like this:

Screen Shot 2020-09-10 at 5 53 04 PM

cc @TylerWanner


zhen0 (approved):
LGTM @znicholasbrown

Issue #215, opened 2020-09-10
Make it more obvious that a user needs to start an agent


jlowin:
Currently, if a user has no agents, they see:
![image](https://user-images.githubusercontent.com/153965/92762190-5794c900-f360-11ea-9ff6-0f31138f9f48.png)

Since this is the common starting point for new users, there's an opportunity to make this more informative and helpful. For example, telling a user that they must start an agent to do work (a common misunderstanding for users that don't read the docs), or jumping-off to actions like creating a new agent config (thanks to the new persistence work by @joshmeek and @cicdw). As is, new users may not realize that this is actually a blocking problem for getting work done (if only due to its yellow color)

In addition, if there is late work AND no agents, that could be surfaced more obviously/visibily than if there's no work and no agents.

cicdw:
Yea, good idea - an easy first fix can change the color + update the message when there are no agents.

znicholasbrown:
@zhen0's PR (#214) will introduce some features to make this more obvious on the upcoming runs and flow page details tiles. Also agreed with @cicdw that we can change the tile color/message to make this more obvious and provide some links to the docs.

Issue #167, opened 2020-08-28
Surface GraphQL errors more directly when performing mutations


znicholasbrown:
**Why**

We currently obfuscate errors that are returned from many of our mutations; this can be convenient for sanitization purposes but isn't particularly helpful when trying to debug, particularly in Server instances.

**What**

* [ ] when any mutation fails, the error message is displayed in a meaningful way (whether it's parsed or the direct message from the API)
* [ ] in locations where real estate is limited (for example in the tenant switcher), errors are surfaced via the `setAlert` store
* [ ] in locations where real estate is not limited, errors are surfaced in the components from which they originate

**Notes**

PR #217, opened 2020-09-11
Labels with agent store

zhen0:
PR Checklist:

## Describe this PR
- Adds a label warning component which checks flow and agent labels and raises an alert on the flow details tile and the upcoming runs tile if there is a mismatch
- Spin off from #214 to use an agent store vs individual agent queries (this slightly goes against recommended practice but after discussion with @znicholasbrown, we decided it might be worth trying to centralize more of our queries rather than making multiple duplicates)
- Also adds a flow store to pass updated label information (although unless there's a slow connection I don't find this any more dynamic than apollo updates)

- Also improves error handling for schedule toggle and label update on the flow page and disabled the add label button for read only users. 
- Updates the agent store unit tests and adds basic flow store unit tests

@jlowin and @cicdw - the upcoming runs gql query is perhaps a little bottom heavy as it queries flow groups from flow runs.  I've not seen any time-outs but would appreciate your review and suggestions if it's likely to be non-performant. 

Should help towards #215 and #167


src/components/LabelWarning.vue
```diff
@@ -0,0 +1,175 @@
...
+        }
+      }
+    }
+  },
+  methods: {
+    labelMessage(message, docsName, link) {
+      this.infoMessage = message
+      this.docsName = docsName
+      this.docsLink = link
+    }
```
znicholasbrown, line 107: I think we could increase the clarity of this component and the clarity and idempotence of the `labelsAlign` prop by moving each of the props you set here to their own computed props whose values are determined by the value of `labelsAlign`. Something like this:

```js
labelsAlign() {
  return !this.flowLabels.length && this.agentLabels.every(labels => !!labels?.length)
},
docsLink() {
  return this.agents?.length ? 'https://docs.prefect.io/orchestration/agents/overview.html#flow-affinity-labels' : 'https://docs.prefect.io/orchestration/agents/overview.html'
}
```
etc. etc.

znicholasbrown (dismissed):
Great progress on this @zhen0, sorry for the slow review. I've made some code design comments and asked some questions; lmk if anything is unclear.

Something I noticed with the design:
Screen Shot 2020-09-16 at 12 02 50 AM
I think showing the green label icon decreases the importance of the error label state you're introducing here, and also makes situations where it doesn't appear look really odd. 

I'd suggest something inline like this (but with better sizing, this feels quite small):
Screen Shot 2020-09-16 at 12 19 34 AM

src/components/LabelWarning.vue
```diff
@@ -0,0 +1,163 @@
...
+            if (this.flowLabels.every(label => array.includes(label)))
+              matchingLabels++
+          })
+        }
+        if (matchingLabels > 0) {
+          this.labelMessage('Your flow and agent labels look good.')
+          return true
+        } else {
+          this.labelMessage(
+            'It looks like you have a mismatch of labels between your flow and running Agents. To allow an Agent to pick up this flow run, you need to have at least one Agent whose labels include those on the flow.'
```
cicdw, line 82:
```suggestion
            'It looks like no currently running Agent has this flow's full set of labels. To allow an Agent to run this flow, you need to have at least one Agent whose labels include all of those on the flow.'
```

cicdw: The wording proposal is hopefully more agnostic to whether the warning is displayed on a flow run or flow tile, and is slightly shorter

src/components/LabelWarning.vue
```diff
@@ -0,0 +1,163 @@
...
+          'You have no live Agents - scheduled flow runs will not be submitted for execution and will display as "Late".'
+        )
+        return false
+      }
+      if (
+        !this.flowLabels?.length &&
+        this.agentLabels.every(arrayOfLabels => arrayOfLabels.length > 0)
+      ) {
+        this.labelMessage(
+          'You have no Agents configured to pick up flows without labels; you may need to add labels to your flow.'
```
cicdw, line 66:
```suggestion
          'You have no currently running Agents configured to pick up flows without labels; you may need to add labels to your flow.'
```

src/components/LabelWarning.vue
```diff
@@ -0,0 +1,163 @@
...
+    },
+    docsLink() {
+      if (!this.agents?.length)
+        return 'https://docs.prefect.io/orchestration/agents/overview.html'
+      return 'https://docs.prefect.io/orchestration/agents/overview.html#flow-affinity-labels'
+    },
+    labelsAlign() {
+      if (!this.agents?.length) {
+        this.labelMessage(
+          'You have no live Agents - scheduled flow runs will not be submitted for execution and will display as "Late".'
```
cicdw, line 57:
```suggestion
          'You have no live Agents - scheduled flow runs will not be submitted for execution and might display as "Late".'
```

Super nit-picky, but there could be situations where a Flow is scheduled far in the future and doesn't have active Agents polling

src/components/LabelWarning.vue
```diff
@@ -0,0 +1,163 @@
...
+          <router-link
+            target="_blank"
+            :to="{
+              name: 'dashboard',
+              params: { tenant: tenant.slug },
+              query: { agents: '' }
+            }"
+          >
+            <span>agents tab</span></router-link
+          ></div
```
cicdw, line 135:
```suggestion
          >.</div
```
in case the diff is hard, this adds a period

src/graphql/Dashboard/upcoming-flow-runs.gql
```diff
@@ -15,6 +15,11 @@ query UpcomingFlowRuns($projectId: uuid) {
       id
       name
       schedule
+      environment
+      flow_group {
+        id
+        labels
```
cicdw, line 21: This should be a fast lookup on a reasonably "small" table, so I expect this will be OK.

This does make me realize that the API does not provide a source of truth for what the "active" labels are, causing the UI to re-implement the logic.  Making a mental note in case there's an easy way to store this more efficiently.

cicdw (changes requested):
This is a game changer and huge UX improvement IMO, very nice work!!

I've left some very minor wording change suggestions for the tooltips.

src/graphql/Dashboard/upcoming-flow-runs.gql
jlowin: I agree, although Hasura's tendency to replace straight joins with nested loops sometimes makes this difficult to predict.

zhen0:
Screen Shot 2020-09-17 at 4 25 17 PM

Updated styling added.  Thanks to Nicholas for his help with the label icon.

zhen0:
But we've encountered a bug with tenant switching and the apollo query not refreshing.  Nicholas is investigating how/if we can do this in an elegant way.

zhen0:
I added cache and reset to test and make sure we could add refresh agents.  Ready for another review @znicholasbrown

znicholasbrown (approved):
Great stuff @zhen0 - LGTM

PR #219, opened 2020-09-11
Provide a way to modify the default server url.

### LINKED ISSUE #99: Provide a way to modify the server url in the built application
**Why**

There's no way to update the default Server url for all users; if the location of the API changes, anyone who visits the UI needs to update the endpoint manually via the Home page. In the previous version of Server we provided a `~/.prefect/config.toml`-level variable, `server.ui.graphql_url`, which ran a small bash script to replace the interpolated endpoint in the built application. The previous solution wasn't ideal because it looked for a specific string (`localhost:4200/graphql`) to replace, meaning the UI images would need to be re-pulled to change the endpoint more than once. A new solution shouldn't have this problem. 

**What**

* [ ] someone deploying the UI can set an environment variable to change the default `server_url` endpoint
* [ ] setting this environment variable will change the default for any users who are served the built application

**Notes**

Here's the previous script we used: 
```bash
  
#!/usr/bin/env bash


if [ "$PREFECT_SERVER__GRAPHQL_URL" != 'http://localhost:4200/graphql' ]
then
    echo "Replacing graphql references with: $PREFECT_SERVER__GRAPHQL_URL"
    for i in /var/www/js/*.js; do
        echo $i
        sed -i -e "s,http://localhost:4200/graphql,$PREFECT_SERVER__GRAPHQL_URL,g" $i
    done
fi

echo "👾👾👾 UI running at localhost:8080 👾👾👾"

nginx -g "daemon off;"

````

### PR DESCRIPTION
**Pull request by znicholasbrown**
PR Checklist:

## Describe this PR
Introduces a new `settings.json` file as a static asset which can be read by the built application. This allows us to inject runtime environment variables to the built application without having to rebuild, meaning variables passed from `docker-compose.yml` when running `prefect server start` with a custom `~/.prefect/config.toml` will be respected again (but better than before).

Resolves #99 

## Notes
This is a tandem PR with PrefectHQ/prefect#3301, which changes the injected variable name.

### DIFF
--- a/src/store/api/index.js
+++ b/src/store/api/index.js
@@ -15,8 +15,8 @@ const state = {
   cloudUrl: process.env.VUE_APP_CLOUD_URL,
   retries: 0,
   serverUrl:
-    settings?.server_url ||
     localStorage.getItem('server_url') ||
+    settings?.server_url ||
     process.env.VUE_APP_SERVER_URL,
   version: null
 }
--- a/src/store/api/index.js
+++ b/src/store/api/index.js
@@ -1,5 +1,6 @@
 import { fallbackApolloClient } from '@/vue-apollo'
 import LogRocket from 'logrocket'
+import settings from '../../../public/settings.json'
 
 const maxRetries = 3
 
@@ -14,7 +15,9 @@ const state = {
   cloudUrl: process.env.VUE_APP_CLOUD_URL,
   retries: 0,
   serverUrl:
-    localStorage.getItem('server_url') || process.env.VUE_APP_SERVER_URL,
+    settings?.server_url ||
+    localStorage.getItem('server_url') ||
+    process.env.VUE_APP_SERVER_URL,
   version: null
 }
 
--- a/src/store/api/index.js
+++ b/src/store/api/index.js
@@ -1,6 +1,5 @@
 import { fallbackApolloClient } from '@/vue-apollo'
 import LogRocket from 'logrocket'
-import settings from '../../../public/settings.json'
 
 const maxRetries = 3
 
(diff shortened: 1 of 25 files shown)

### CODE REVIEW

File: src/store/api/index.js [hunk 1]
```diff
@@ -14,7 +15,9 @@ const state = {
   cloudUrl: process.env.VUE_APP_CLOUD_URL,
   retries: 0,
   serverUrl:
-    localStorage.getItem('server_url') || process.env.VUE_APP_SERVER_URL,
+    settings?.server_url ||
```
zhen0, line 18: I wonder about the ordering here? If I change the serverUrl in the home page, it updates but then when I refresh the page it goes back to the one set in settings. Is there a reason why it checks settings before local storage?

zhen0 (commented):
One comment/question about how this interacts with localstorage/the url set in the UI homepage @znicholasbrown

src/store/api/index.js
znicholasbrown: Hm nope, you're right, that's definitely the incorrect oder. Fixed!

src/store/api/index.js
```diff
@@ -1,5 +1,6 @@
 import { fallbackApolloClient } from '@/vue-apollo'
 import LogRocket from 'logrocket'
+import settings from '../../../public/settings.json'
```
znicholasbrown, line 3: Something's up with this import, I don't think it's being parsed as a static asset - investigating.

znicholasbrown:
Ok this should be ready to go again. The way to test this locally:
```
$ docker build prefecthq/ui:latest
# on an editable install of the master branch of Core
$ prefect server start --skip-pull
```

If you've worked with Server previously, you may have a `localStorage` key set; you can clear this in the dev console with `localStorage.clear()`, and then refresh the Server UI.

joshmeek:
After testing can confirm that this works!

joshmeek (approved):

PR #227, opened 2020-09-16
Clear late runs

cicdw:
PR Checklist:

## Describe this PR
This PR fixes a minor bug that prevented users from clearing late runs from many of the dashboard pages.  In particular, it seemed our "Clear" button was also attempting to update all task run states to "Cancelled".  On some pages, this piece of the mutation generation was erroring out because there were no referenceable task runs around.

While this is nice, it's not necessary and also could result in performance issues for flows with 100s or 1,000s of tasks.  This PR only updates the flow run state to "Cancelled", which conveniently resolves the bug we were seeing.


zhen0 (approved):
This works a lot more smoothly and looks good to me @cicdw.  At some point we should standardize the dialog and buttons across the dashboard and flow page but that's for another PR.

znicholasbrown (approved):

PR #232, opened 2020-09-17
Performance work (part...3?)

znicholasbrown:
PR Checklist:

## Describe this PR
* Improves error handling related to garbage collection
* Improves component teardown by accounting for elements whose `destroyed` method had been called but that hadn't been removed from the DOM due to transitions
* Improves component teardown by expanding the keys we're resetting (including the apollo vm), as well as adding a general loop for non-standard keys
* Fixes some errors that were thrown when routing to tenant-guarded pages without passing a tenant slug
* Navigating back or forward after switching the tenant now forces an update of the dashboard (and resets the current tenant as expected)
* General routing improvements that should make navigating with the forward and back buttons smoother and more consistent

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -5,11 +5,13 @@
 ### Features and Improvements
 
 - Add unit tests for tenant store [#195](https://github.com/PrefectHQ/ui/pull/195)
-- None
+- Improves error handling related to garbage collection - [#232](https://github.com/PrefectHQ/ui/pull/232)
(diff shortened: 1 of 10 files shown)


zhen0 (dismissed):
This is looking sooo much better.  And I have a clear console.  Thanks for all your work on this @znicholasbrown

zhen0 (approved):
LGTM @znicholasbrown

PR #241, opened 2020-09-21
Apollo endpoint usability update

znicholasbrown:
PR Checklist:

## Describe this PR
* Removes `BASE_URL` from the environment variables file
* Fetch `prefect_ui_settings` static JSON from `/` instead of the base url
* Add a reset method to the GraphQL endpoint setter input, which will clear the `localStorage` `server_url` key and reset default URL to whatever was configured either from the environment or the settings.json

Should make deploying to a remote server with a non-static domain easier

### DIFF
--- a/src/pages/Home/StartPrefectServer-Section.vue
+++ b/src/pages/Home/StartPrefectServer-Section.vue
@@ -34,6 +34,10 @@ export default {
 
       this.urlInput = this.defaultUrl
 
+      this.error = false
+      this.success = false
+      this.connected = false
+
       this.setServerUrl(this.defaultUrl)
     },
     async _testUrl() {
@@ -156,7 +160,7 @@ export default {
           </template>
 
           <template v-slot:append-outer>
-            <div class="mt-n1">
+            <div>
               <v-btn
                 color="primary"
                 dark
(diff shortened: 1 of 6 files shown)


src/pages/Home/StartPrefectServer-Section.vue
```diff
@@ -114,14 +124,32 @@ export default {
...
+            <v-tooltip top>
+              <template v-slot:activator="{ on }">
+                <v-btn
+                  color="blue-grey lighten-1"
+                  icon
+                  small
+                  @click="_resetUrl"
+                  v-on="on"
+                >
+                  <v-fade-transition mode="out-in">
```
zhen0, line 136: I find the mix of indicator and action button a bit confusing here.  Could we keep the indicator as is and add a reset button next to the connect button?  (Or something similar.)

zhen0 (changes requested):
I like the addition of the rest but one suggestion/request @znicholasbrown

znicholasbrown:
Good call @zhen0 - that's been updated:
Screen Shot 2020-09-23 at 11 24 23 AM

zhen0 (approved):
Thanks for the changes @znicholasbrown

PR #258, opened 2020-09-28
Show user notifications on the notifications tile

znicholasbrown:
PR Checklist:

## Describe this PR
^^ and also add dense styling for the WhatsNew notification component
Screen Shot 2020-09-28 at 3 33 10 PM

### DIFF
--- a/src/pages/Dashboard/Notifications-Tile.vue
+++ b/src/pages/Dashboard/Notifications-Tile.vue
@@ -51,7 +51,10 @@ export default {
     },
     where() {
       let where = {
-        tenant_id: { _eq: this.tenant?.id },
+        _or: [
+          { tenant_id: { _eq: this.tenant?.id } },
+          { tenant_id: { _is_null: true } }
+        ],
         read: { _eq: false }
       }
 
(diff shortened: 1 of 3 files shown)


src/pages/Dashboard/Notifications-Tile.vue
```diff
@@ -51,7 +51,10 @@ export default {
     },
     where() {
       let where = {
-        tenant_id: { _eq: this.tenant?.id },
+        _or: [
```
cicdw, line 54: is this an exclusive or?

cicdw: I'm wondering about the case where a user has multiple tenants and this runs
```
{ tenant_id: { _eq: null } },
{ tenant_id: { _is_null: true } 
```

znicholasbrown: I think the query will fail in that specific case as you're expecting, since `== null && !== null` can't both be true. But a whole lot else will be failing if `tenant.id == null` so I'm not super worried about that here.

cicdw: @znicholasbrown Hasura matches _everything_ when you use `_eq: null` (hence the `_is_null` special comparison)

cicdw (approved):

src/pages/Dashboard/Notifications-Tile.vue
znicholasbrown: This was resolved offline^

PR #266, opened 2020-09-29
Auth nav unit tests

### LINKED ISSUE #150: Unit test for authnavguard middleware
**Why**
Increase our test coverage and give us increased confidence in the auth nav guard/middleware

**What**
* [ ] Test all possible types of from routes 
* [ ] Test all possible types of to routes
* [ ] Test all possible authentication types

**Notes**
Sub issue of #101 
Some tests and template/examples in cloud-web-ui

zhen0:
PR Checklist:

## Describe this PR
- Adds unit tests for the auth nav guard middleware
- Closes #150


tests/unit/middleware/authNavGuard.spec.js
```diff
@@ -0,0 +1,205 @@
...
+    })
+
+    const next = jest.fn()
+
+    expect(store.getters['auth0/isAuthorized']).toBe(false)
+
+    await authNavGuard({}, {}, next)
+
+    expect(store.getters['auth0/isAuthorized']).toBe(true)
+
```
znicholasbrown, line 122: Same comment as above re: adding an assertion that `auth0/authorize` is called

tests/unit/middleware/authNavGuard.spec.js
```diff
@@ -0,0 +1,205 @@
...
+      store.commit('user/user', { name: 'test_user' })
+    })
+
+    const next = jest.fn()
+    expect(store.getters['user/userIsSet']).toBe(false)
+
+    await authNavGuard({}, {}, next)
+
+    expect(store.getters['user/userIsSet']).toBe(true)
+
```
znicholasbrown, line 143: Same comment as above re: adding an assertion that `user/getUser` is called

znicholasbrown (dismissed):
Just a few comments but all the same tweak @zhen0

zhen0:
Thanks for the comments @znicholasbrown - I haven't added a check for 'authenticate' or 'authorize' being called because I could not get spying on store.dispatch or store.actions to work.  I've updated the wording of the tests and added comments to reflect what we are testing and also added an issue in #297 to come back and to figure out that spy. 

I've also updated the tenant store unit tests to remove a mockError that was causing a pre-commit hook error for @whimsicallyson - standard git commit should work for now.

znicholasbrown (approved):
Approved with the new changes 👍

PR #269, opened 2020-09-29
Remove "A nonexistant user" from flow and flow run details for server instances

joshmeek:
PR Checklist:

## Describe this PR

Currently all flows and subsequent runs for server users show that they are created by `A nonexistant user` and this is not helpful or could be confusing to users of server because the backend has no concept of users. This PR updates the flow details tile to only display this section if a user is attached to a flow and it updates the text output of the flow run details tile.

**Flow**

Before:
image

After:
image

**Flow Run**

Before:
image

After:
image

**Note** this might not be the best wording for an ad hoc / manually created run. Open to suggestions!

### DIFF
--- a/src/pages/Flow/Details-Tile.vue
+++ b/src/pages/Flow/Details-Tile.vue
@@ -98,7 +98,7 @@ export default {
       )
     },
     hasUser() {
-      return this.flow.created_by != null
+      return this.flow?.created_by
     }
   },
   methods: {
@@ -274,9 +274,8 @@ export default {
               <div v-if="hasUser" class="subtitle-2">
                 {{ flow.created_by.username }}
               </div>
-              <div class="caption">
-                <b v-if="!hasUser"> {{ formatTime(flow.created) }} </b>
-                <span v-if="hasUser"> {{ formatTime(flow.created) }} </span>
+              <div class="caption" :class="{ 'font-weight-bold': !hasUser }">
+                {{ formatTime(flow.created) }}
               </div>
             </v-list-item-content>
           </v-list-item>
--- a/src/pages/Flow/Details-Tile.vue
+++ b/src/pages/Flow/Details-Tile.vue
@@ -266,15 +266,18 @@ export default {
     <v-card-text class="pl-12 pt-2 card-content">
       <v-fade-transition hide-on-leave>
         <div v-if="tab == 'overview'">
-          <v-list-item v-if="hasUser" dense class="px-0">
+          <v-list-item dense class="px-0">
             <v-list-item-content>
               <v-list-item-subtitle class="caption">
-                Created by
+                Created <span v-if="hasUser">by</span>
               </v-list-item-subtitle>
-              <div class="subtitle-2">
+              <div v-if="hasUser" class="subtitle-2">
                 {{ flow.created_by.username }}
               </div>
-              <div class="caption">{{ formatTime(flow.created) }} </div>
+              <div class="caption">
+                <b v-if="!hasUser"> {{ formatTime(flow.created) }} </b>
+                <span v-if="hasUser"> {{ formatTime(flow.created) }} </span>
+              </div>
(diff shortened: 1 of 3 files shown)


src/pages/Flow/Details-Tile.vue
```diff
@@ -266,16 +269,15 @@ export default {
...
-                  flow.created_by
-                    ? flow.created_by.username
-                    : 'A nonexistent user'
-                }}
+              <div v-if="hasUser" class="subtitle-2">
+                {{ flow.created_by.username }}
+              </div>
+              <div class="caption">
+                <b v-if="!hasUser"> {{ formatTime(flow.created) }} </b>
+                <span v-if="hasUser"> {{ formatTime(flow.created) }} </span>
```
joshmeek, line 279: Not sure what the best way is to _only_ bold this if there is no user

src/pages/FlowRun/Details-Tile.vue
```diff
@@ -146,7 +146,7 @@ export default {
                     ? 'Prefect Scheduler'
                     : flowRun.created_by
                     ? flowRun.created_by.username
-                    : 'A nonexistant user'
+                    : 'Ad hoc run'
```
joshmeek, line 149: Open to better wording suggestions

cicdw: Random ideas:
- external process
- API call
- POST request
- just the word "user" (<-- this one best matches the alternate "Prefect Scheduler")

joshmeek: Yeah I kind of like _user_/_User_ since it is literally created by _a_ user lol

znicholasbrown: What about removing this field entirely for Server? Having "User" suggests that user as a concept exists in the system, which it doesn't.

znicholasbrown:
I'm in favor of removing this line item for Server but not from Cloud. The existing message is useful when users are removed from the system entirely; it suggests this _was_ created by a user, they just don't exist anymore.

src/pages/FlowRun/Details-Tile.vue
joshmeek: The reason I left it in is because I like that we show that the run was auto scheduled with the Prefect Scheduler and something different for if it was created manually. I don't feel too strongly about it and am fine for removing it though

znicholasbrown: Yeah I think auto-scheduled information is helpful so definitely in favor of keeping that but hiding the user portion for Server

src/pages/Flow/Details-Tile.vue
zhen0, line 279: I'd alter the class based on the hasUser property.  Something like this: 
```suggestion
             <div class="caption" :class="{ 'font-weight-bold': !hasUser }">
                {{ formatTime(flow.created) }}
              </div>
```

zhen0 (changes requested):
Nice update @joshmeek! A couple of small suggestions/questions and I agree with Nicholas's comment that we should keep created_by 'non existent user' for Cloud.

src/pages/Flow/Details-Tile.vue
zhen0: For some reason it's shown up separately but see my suggestion below.

src/pages/Flow/Details-Tile.vue
```diff
@@ -96,6 +96,9 @@ export default {
           (val, index) => val === this.flow?.environment?.labels[index]
         )
       )
+    },
+    hasUser() {
+      return this.flow.created_by != null
```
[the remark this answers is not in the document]
joshmeek: No reason at all! Learning the syntax as I go

znicholasbrown (approved):
LGTM @joshmeek 👍

PR #273, opened 2020-09-30
Allow label editing on older versions of core and disable empty entries

zhen0:
PR Checklist:

## Describe this PR
-Fixes an issue where flows registered with older versions of core (and so had no labels) could not add flow group labels from the UI
- As raised in https://prefect-community.slack.com/archives/C0192RWGJQH/p1601436207001800

### DIFF
--- a/src/pages/Flow/Details-Tile.vue
+++ b/src/pages/Flow/Details-Tile.vue
@@ -83,7 +83,8 @@ export default {
       const labels =
         this.newLabels ||
         this.flowGroup?.labels ||
-        this.flow?.environment?.labels
+        this.flow?.environment?.labels ||
+        []
       return labels?.slice().sort()
     },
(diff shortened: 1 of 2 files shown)


src/pages/Flow/Details-Tile.vue
```diff
@@ -83,7 +83,8 @@ export default {
       const labels =
         this.newLabels ||
         this.flowGroup?.labels ||
-        this.flow?.environment?.labels
+        this.flow?.environment?.labels ||
+        []
```
cicdw, line 87: Ahh this makes sense! Nice catch

cicdw (approved):
LGTM - nice catch! I originally thought this was going to be a browser issue

znicholasbrown (approved):

PR #278, opened 2020-10-01
Timeline

znicholasbrown:
PR Checklist:

## Describe this PR
This is a draft PR for the upcoming flow run timeline feature, to replace the existing gantt-chart on the flow run page. 

This is a work in progress but I'm opening the PR up to comments/requests. Before making a request, make sure that your request isn't already covered in the list below.

- [x] remove apexcharts
- [x] improve collapsed view (currently unused but available as a `Timeline` component prop); perhaps use the bar outline for opacity-based visualization, this view breaks with lots of parallel runs or a small scale with lots of tasks (300+)
- [x] move tasks without start times to a "staging area" in the future portion of the timeline
- [x] allow live updating when a flow run is underway
- [x] introduce breakpoints for important flow run-level events (like when the flow run is scheduled, submitted, or running)
- [x] add hover interactions to the canvas elements (component)
- [x] add hover interactions to the breakpoints (component)
- [x] show information about breakpoints on hover (tile-level)
- [x] show information about task run on hover (tile-level)
- [x] show task labels
- [x] task labels should always be visible when a bar is in the current viewport
- [ ] add click interactions to the canvas elements (component)
- [ ] do **something** ?? when a task run bar is clicked, perhaps navigate to that task run page
- [x] implement segmented bars for mapped task runs
- [x] use the new `mapped_children` query route to get mapped task run states
- [x] use the new `no_mapped_states` to dictate if a mapped task is still running for newer versions of Core
- [x] rename existing timeline components
- [x] break existing max-width restrictions
- [x] display a breakpoint on hover that gives the timestamp of the mouse location
- [ ] ... also show the number of running tasks at a given hover point
- [x] rename tabs and tiles (remove gantt, add timeline)
- [x] improve the design of the x-axis; I have some sketches but they'll require updating the x-scale date parsing
- [x] improve flow run page transitions (for performance and consistency we'll use fade animations instead of slide)
- [ ] implement a brush chart
- [x] add optional breakpoint for the `Cancelling` flow run state transition
- [x] add bar alpha transitions
- [ ] make timestamps timezone aware
- [x] start breakpoint transitions from the right (future) instead of the left


src/components/TaskRunMenu.vue
```diff
@@ -0,0 +1,106 @@
+<script>
```
zhen0, line 1: Not an issue just intrigued as I can't see it - where are we using this component?

zhen0 (approved):
This is looking great @znicholasbrown  Tested with a few different flows and it's very cool to watch them running.  One very small question and one idea for the future - I would love if we could click on a task run from the tile and get taken to that task run's page or have a solid popover info box with a link to the task run. 

I'll approve but leave you to merge in case you want reviews from others.

src/components/TaskRunMenu.vue
znicholasbrown: Ah yes - this component is related to your comment about a more robust popover; I haven't adapted it to the new data structure of the timeline but it's used in the gantt chart right now.

zhen0 (approved):
Looks great @znicholasbrown

PR #285, opened 2020-10-02
Restart

### LINKED ISSUE #226: Abililty to pause/resume flow_run after cancelling
**Why**
Our teams need abililty to pause/resume flow runs without affecting other flow runs for the same flow. So multiple teams can run the flow in parallel and have ability to control their run without affecting other teams. 

Currently when we cancel a flow run, it cannot be resumed (only failed ones are allowed to be restarted. Can we allow ability resume cancelled flow runs?
**What**
I want the ability to resume running a Flow Run once I cancelled it. I need the abililty to pause/resume a Flow run independent of the other flow runs for that Flow. .

* [ ] 

**Notes**
For now I can change the state of a flow run to "Paused" and when we need to resume, change to "Scheduled" so agent can pick up and run. Will that do what I am looking for?

### LINKED ISSUE #283: Update restart from failed to include restart from cancelled
**Why**

Update the restart button to allow restarting from cancelled (and maybe cancelling) states. 
Slack discussion: https://prefecthq.slack.com/archives/CJN6L95L7/p1600221814038000
In part would meet this request #226 

**What**

* [ ] Restart from cancelled is allowed from the UI
* [ ] The UI OR Cloud correctly handles task runs that are still running even if the task has reached cancelled - perhaps by setting state to cancelled/failed or finished
* [ ] Restart from cancelling is allowed (but as that's a running state this could be more complicated?)

**Notes**

zhen0:
PR Checklist:

## Describe this PR
- Adds cancelled as a state from which a user can restart a flow run.  This means that if a user cancels a flow run, they can use restart to re-run any task runs in a cancelled or failed state and to run any task runs still in a pending state.  
- After discussion with @jcrist (who did a lot of work to implement cancellation), this PR does does not include cancelling as it's still a running state and is basically waiting for all running task runs to enter a finished state.  
- I've also updated the disabling to be done from the action bar button rather than the dialog so it's a bit more user friendly and included more information in the tooltip.  
- Also fixed a bug with the restart from task run button (which was passing an incorrectly formatted variable). 
- Closes #283 and closes #226


znicholasbrown (dismissed):
This is working nicely @zhen0 - just a couple wording comments and a missing / incorrect tooltip on the confirm buttons for both dialogs:
Screen Shot 2020-10-06 at 1 29 02 PM

zhen0:
Thanks @znicholasbrown - some really useful comments.  I've removed the tooltips and updated the wording about restart further.

znicholasbrown (approved):
LGTM @zhen0 👍

PR #295, opened 2020-10-06
Apollo url speed racer

znicholasbrown:
PR Checklist:

## Describe this PR
Moves the server ui settings out of the index head and into `main.js` as a pre-condition to initializing the Vue app. This should remove the race condition which caused the `apollo_url` setting to sometimes be ignored.

### DIFF
--- a/src/store/api/index.js
+++ b/src/store/api/index.js
@@ -1,6 +1,8 @@
 import { fallbackApolloClient } from '@/vue-apollo'
 import LogRocket from 'logrocket'
 
+const UI_DEPLOY_TIMESTAMP = process.env.VUE_APP_RELEASE_TIMESTAMP
+
 const maxRetries = 3
 
 const state = {
@@ -14,7 +16,7 @@ const state = {
   cloudUrl: process.env.VUE_APP_CLOUD_URL,
   retries: 0,
   serverUrl:
-    localStorage.getItem('server_url') ||
+    localStorage.getItem(`${UI_DEPLOY_TIMESTAMP}_server_url`) ||
     window.prefect_ui_settings?.server_url ||
     process.env.VUE_APP_SERVER_URL,
   version: null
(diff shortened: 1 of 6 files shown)


tests/unit/store/api.spec.js
```diff
@@ -101,7 +103,7 @@ describe('API Vuex Module', () => {
         cloudUrl: process.env.VUE_APP_CLOUD_URL,
         retries: 5,
         serverUrl:
-          localStorage.getItem('server_url') || process.env.VUE_APP_SERVER_URL,
+          localStorage.getItem(SERVER_KEY) || process.env.VUE_APP_SERVER_URL,
```
zhen0, line 106: Should we keep this the same as the actual api store state serverUrl and take out the process.env.VUE_APP_SERVER_URL?

src/store/api/index.js
```diff
@@ -1,6 +1,15 @@
 import { fallbackApolloClient } from '@/vue-apollo'
 import LogRocket from 'logrocket'
 
+const SERVER_KEY = `${process.env.VUE_APP_RELEASE_TIMESTAMP}_server_url`
```
zhen0, line 4: Not a big deal but will this end up filling local storage with lots of unused keys?

zhen0 (commented):
A few questions @znicholasbrown - not necessarily changes

src/store/api/index.js
znicholasbrown: I think it should be ok - this can always be cleared by a user (as part of a broader browser cache clear) but if not we use localStorage sparingly enough elsewhere that this shouldn't have an impact. Good callout though, let's keep an eye on it and find a more permanent solution if needed.

tests/unit/store/api.spec.js
znicholasbrown: We're still falling back on the environment variable (line 9 of the api store index) so I don't think we should remove unless there's a testing reason to do so.

zhen0: Ok - that sounds fair.

zhen0 (approved):
LGTM @znicholasbrown

PR #302, opened 2020-10-08
Display task run names on flow run page and task run page

### LINKED ISSUE #301: Display task run names
**Why**
Now that https://github.com/PrefectHQ/prefect/pull/3420 has been released we should display task run names when set

**What**
- [ ] Display task run names on task run page
- [ ] Display task run names on flow run page for table of task runs
- [ ] Display task run names in activity tile on flow run page
- [ ] Determine if there are any other relevant places where they should be displayed

**Notes**
N/A

joshmeek:
PR Checklist:

## Describe this PR

Closes #301 

This PR exposes the new task run names if they are set. The task run names on the flow run page and task run page are updated (screenshots below). Here is the flow code used if anyone wants to reproduce:

```python

@task
def get_values():
    return ['my_value', 'another_one', 'nicholas', 'marvin']

@task(log_stdout=True, task_run_name=lambda **kwargs: f"{kwargs['v']}")
def compute(v):
    print(v)

with Flow("task_run_names") as flow:
    vals = get_values()
    compute.map(vals)

flow.register(project_name="Demo")
```

_[Flow Run Page]_ Current vs New 

image

_[Task Run Page]_ Current vs New

image

Let me know if there are any more places we should expose this / if my boolean JS logic is correct 🙂

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -6,6 +6,7 @@
 ### Features and Improvements
 
 - Add unit tests for the authNavGuard middleware [#266](https://github.com/PrefectHQ/ui/pull/266)
+- Display task run names on flow run page and task run page [#302](https://github.com/PrefectHQ/ui/pull/302)
 
 ### Bugfixes
 
--- a/src/components/HeartbeatTimeline.vue
+++ b/src/components/HeartbeatTimeline.vue
@@ -77,7 +77,7 @@ export default {
           name = item.name
           break
         case 'task_run':
-          name = item.task.name
+          name = item.name || item.task.name
           break
         case 'task_run_state':
           name = item.message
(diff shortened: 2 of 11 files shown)


zhen0:
Thanks @joshmeek - this is looking good.  I think one other place where it would be useful to have task run names is the schematics.  At present we list mapped children but without any identifying features - could we include names in there?  (Happy to pair on that if digging into schematics get touch.)
Screen Shot 2020-10-08 at 12 15 13 PM

zhen0 (approved):
LGTM @joshmeek - but could you update master

znicholasbrown (approved):
🎉

PR #308, opened 2020-10-08
Invalid state hotfix

znicholasbrown:
PR Checklist:

## Describe this PR
Adds a try catch to the auth redirect callback to catch repeated calls to that method.

### DIFF
--- a/src/store/auth0/index.js
+++ b/src/store/auth0/index.js
@@ -222,29 +222,27 @@ const actions = {
   async authenticate({ dispatch, commit, getters }) {
     if (!auth0Client) await dispatch('initAuth0')
 
-    if (window.location.search.includes('code=')) {
+    if (window.location?.search?.includes('code=')) {
       try {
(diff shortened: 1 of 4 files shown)


src/store/auth0/index.js
```diff
@@ -223,7 +223,18 @@ const actions = {
...
+      } catch (e) {
+        // Catches repeated calls to the handleRedirectCallback...
+        // The first time handleRedirectCallback is called, it removes the challenge
+        // from the window search... when another call is made, it looks for the challenge
+        // which was removed, halting further execution and navigation. This allows the method to proceed
+        // in cases of an invalid state
+        if (!e?.includes('Error: Invalid State')) {
+          return await dispatch('login')
+        }
+      }
```
znicholasbrown, line 237: @zhen0 does this make sense to test? I'm thinking maybe not because we mock the auth0 calls where this error would be thrown.

zhen0 (approved):

PR #323, opened 2020-10-13
impose height in last ten runs div to keep row consistency

### LINKED ISSUE #289: BUG: Flow table rows expand past their set heights on load
**Bug Description**

When first visiting the `Flows` tab, the rows break their set heights and are briefly expanded

**Steps To Reproduce**

1. Start from the dashboard
2. Visit the `Flows` tab

**Browsers Tested:**
* [x] Chrome
* [ ] Firefox
* [ ] Safari
* [ ] Edge
* [ ] IE

----------
**Notes**

Oct-05-2020 10-21-52

zhen0:
PR Checklist:

## Describe this PR
- Adds a height to the last ten runs div on the flow table tile and the flow versions tile to prevent row height flicker on load
- Closes #289

### DIFF
--- a/src/pages/Dashboard/FlowTable-Tile.vue
+++ b/src/pages/Dashboard/FlowTable-Tile.vue
@@ -375,7 +375,7 @@ export default {
         </template>
 
         <template #item.flow_runs="{ item }">
-          <div class="position-relative allow-overflow">
+          <div class="position-relative allow-overflow" style="height: 55px;">
             <LastTenRuns :flow-id="item.id" :archived="item.archived" />
           </div>
         </template>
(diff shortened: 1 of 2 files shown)


whimsicallyson:
this makes sense to me. do we have standards on when to add a class name vs when to use an inline style?

znicholasbrown:
@whimsicallyson - a loose standard is to use a class if you're using the inline style more than once or you're adding more than one inline style.

znicholasbrown (approved):

whimsicallyson:
could probably also put the style on the surrounding `div` in the `LastTenRuns` component so we don't have to worry about this happening if we want to add another use of it

zhen0:
@whimsicallyson - I wondered about that - there may be cases in future where we want to use the LastTenRuns component in a different way so I went with keeping it more specific.

Issue #319, opened 2020-10-12
BUG: Inconsistent Styling


dylanbhughes:
**Bug Description**

Running into this in our staging environment:
![Screen Shot 2020-10-12 at 2 13 25 PM](https://user-images.githubusercontent.com/2325367/95778971-2c6f0380-0c97-11eb-95b6-3ac3c6a47f17.png)

**Steps To Reproduce**

Unsure, only encountering in our staging environment starting today.

**Browsers Tested:**
* [X] Chrome
* [ ] Firefox
* [ ] Safari
* [ ] Edge
* [ ] IE

----------
**Notes**

raphael-riel:
Got the same issue. Reloading the Browser's page fixes the problem until a click somewhere break the layout again.

MacOS with Chrome 86.

zhen0:
I couldn't recreate this bug but I believe we solved it with #330 which was released on October 21.  @dylanbhughes and @raphael-riel have you seen the inconsistent styling at all in the last week?

dylanbhughes:
@zhen0 I haven't this week

raphael-riel:
@zhen0 All good here too! 👍

PR #330, opened 2020-10-14
Styling

zhen0:
PR Checklist:

## Describe this PR
- Moves styling from the dashboard tile to the new label edit section.  Also makes that styling scoped and more specific. 
- This MAY solve #319 - In the screenshots all the affected areas are list items and the un-scoped styling was on a list-item-content so may have affected them.  However, I can't recreate that styling issue so not 100% confident it's fixed.

### DIFF
--- a/src/components/LabelEdit.vue
+++ b/src/components/LabelEdit.vue
@@ -408,8 +408,4 @@ export default {
 .scroll-across {
   overflow: scroll;
 }
-
-// .v-list-item__content {
-//   overflow: scroll;
-// }
 </style>
--- a/src/components/LabelEdit.vue
+++ b/src/components/LabelEdit.vue
@@ -165,7 +165,7 @@ export default {
 
 <template>
   <v-list-item dense class="px-0">
-    <v-list-item-content width="800px" class="overflow-x-scroll scroll-across">
+    <v-list-item-content width="800px" class="scroll-across">
       <v-list-item-subtitle class="caption">
         Labels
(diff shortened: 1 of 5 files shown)


src/components/LabelEdit.vue
```diff
@@ -165,7 +165,7 @@ export default {
 
 <template>
   <v-list-item dense class="px-0">
-    <v-list-item-content width="800px" class="overflow-x-scroll">
+    <v-list-item-content width="800px" class="overflow-x-scroll scroll-across">
```
znicholasbrown, line 168: I'm not sure `overflow-x-scroll` exists in our stylesheet (it should); if it did, wouldn't this just need the 1 class?

znicholasbrown (commented):
Just 1 comment, I can't reproduce #319 but have seen it on multiple other screens so probably need to get someone else to take a look at this to confirm.

src/components/LabelEdit.vue
zhen0: It didn't but it does now - good call.

znicholasbrown (approved):
This LGTM and I think you're right that this will resolve #319 - @dylanbhughes and I did some digging this afternoon and removing `overflow: scroll` on the item content class seemed to fix the problem. The reason this is inconsistently applied is because the order in which the stylesheets are loaded will determine when that particular class would be applied globally. Either way let's keep an eye on that and re-open #319 if needed once this is released.

PR #331, opened 2020-10-14
Router Bug Fix

### LINKED ISSUE #324: BUG: Occasional "[vue-router] relative params navigation requires a current route." error with blank screen
**Bug Description**

Occasionally (about 1 in 10 page reloads) I'm seeing a blank page when I reload the page (by clicking on the 'reload this page' button next to the URL bar in chrome) .  The top navbar shows but no dashboard or home page.  The console shows a "[vue-router] relative params navigation requires a current route" error/info message.

**Steps To Reproduce**

1. I'm not seeing this consistently.  For me, reload the page many times and you should eventually see it. Possibly coming from some sort of race condition?  

**Browsers Tested:**
* [x] Chrome
* [ ] Firefox
* [ ] Safari
* [ ] Edge
* [ ] IE

----------
**Notes**

Seen in dev and staging. 
Screen Shot 2020-10-13 at 5 47 07 PM

zhen0:
PR Checklist:

## Describe this PR
- Updates the 'finally' section in App.vue to only add a tenant to the route if headed to the dashboard.  This reduces the relative params navigation error that was causing an occasional blank screen when trying to navigate via a URL. 
- There's still a long loading/auth pattern happening that we should handle in a separate PR
- Closes #324

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -190,7 +190,6 @@ export default {
     },
     async startup() {
       this.startupHasRun = true
-
       try {
         if (this.isCloud) {
           if (!this.isAuthorized) {
@@ -240,7 +239,7 @@ export default {
         }
       } finally {
         this.shown = true
-        if (this.isCloud && !this.$route.name && !this.$route.params.tenant) {
+        if (this.route === 'dashboard' && !this.$route.params.tenant) {
           this.$router.replace({
             params: {
               tenant: this.tenant.slug
@@ -292,13 +291,6 @@ export default {
 <template>
   <v-app class="app">
     <v-main :class="{ 'pt-0': isWelcome }">
-      <v-progress-linear
-        absolute
-        :active="isLoggingInUser"
-        indeterminate
-        height="5"
-      />
-
       <v-slide-y-transition>
         <NavBar v-if="shown && loadedComponents > 0" />
       </v-slide-y-transition>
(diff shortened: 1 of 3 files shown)


src/App.vue
```diff
@@ -240,8 +240,7 @@ export default {
         }
       } finally {
         this.shown = true
-
-        if (!this.$route.params.tenant) {
+        if (this.isCloud && !this.$route.name && !this.$route.params.tenant) {
```
znicholasbrown, line 243: Hm I don't think this is correct; this shouldn't be restricted to Cloud (tenants exist in Server now so we still want the slug), and the Dashboard is a named route, aliased to root:
```js
// router.js
 {
    name: 'dashboard',
    alias: '/',
    path: '/:tenant?',
    component: () =>
      import(
        /* webpackChunkName: "dashboard" */ '@/pages/Dashboard/Dashboard.vue'
      ),
    beforeEnter: multiguard([authNavGuard, tenantNavGuard])
  }

```

znicholasbrown (changes requested):
If this fixes the issue the error you're seeing with a blank dashboard then we can confirm this is the correct location but not quite the right fix. This won't push the slug for Server users and I'm not sure when the condition of the unnamed route will be met.

zhen0:
I've updated this @znicholasbrown but still not sure about that router.replace.  The relative params navigation error seems to be fixed and I don't see any missing slug errors but let me know if you see any issues with it. (I did see an occasional missing slug error when switching from server - I've updated the sidenav to deal with that). 

I also added a first iteration of a loading message as I saw a blank screen too many times.  We should build on that but hopefully this can provide users with a sense that their page is loading.

PR #354, opened 2020-10-20
Whitescreen bugfix

znicholasbrown:
PR Checklist:

## Describe this PR
Removes the startup method from the app component in favor holding off on rendering until authentication is finished. This should fix the long wait time, as well as the white screen bug, which arose from a race condition where the router wasn't initialized before we were attempting to push a team slug. 

After some shallow time to paint testing, I was seeing an average of ~4 seconds to get a full dashboard, which doesn't seem unreasonable. Only a fraction of that was actual rendering; the largest contributor are the necessary network checks we do before initialization: the auth nav guard, which handles authentication with auth0 and authorization with Cloud took 2.2 seconds on average, and was the largest contributor overall. Next was the tenant nav guard, taking 1 second on average. The largest contributing method to the above is the `createAuth0Client` call, which we do just once when the application loads, but took sometimes as long as 1 second to complete. Unfortunately this is out of our control for the time being but I'll keep an eye on this metric as we look at other identity providers. For now, I don't think we can eke out much more in terms of performance for startup since each call we make needs to happen in sequence (authenticate => authorize => tenant).

cc @zhen0 @whimsicallyson would you mind testing this? I know you both were able to reproduce some fairly long wait times.

### DIFF
--- a/src/pages/Dashboard/Agents-Tile.vue
+++ b/src/pages/Dashboard/Agents-Tile.vue
@@ -146,7 +146,7 @@ export default {
         </v-list-item>
 
         <v-list-item
-          v-if="agentTracker.healthy > 0"
+          v-if="agentTracker && agentTracker.healthy > 0"
           key="healthy-agents"
           color="grey"
         >
@@ -169,7 +169,7 @@ export default {
         </v-list-item>
 
         <v-list-item
-          v-if="agentTracker.stale > 0"
+          v-if="agentTracker && agentTracker.stale > 0"
           key="stale-agents"
           color="grey"
         >
@@ -195,7 +195,7 @@ export default {
         </v-list-item>
 
         <v-list-item
-          v-if="agentTracker.unhealthy > 0"
+          v-if="agentTracker && agentTracker.unhealthy > 0"
           key="unhealthy-agents"
           color="grey"
         >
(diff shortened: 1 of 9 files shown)


src/pages/Dashboard/Agents-Tile.vue
```diff
@@ -169,7 +169,7 @@ export default {
         </v-list-item>
 
         <v-list-item
-          v-if="agentTracker.stale > 0"
+          v-if="agentTracker && agentTracker.stale > 0"
```
zhen0, line 172: Guessing this removes the console errors - nice!

zhen0 (dismissed):
LGTM @znicholasbrown - I still see an average loading time of about 6 seconds to a full screen but it's a definite improvement.  The progress also feels more tangible with the nav bar showing/loading much sooner.

znicholasbrown:
Ok great - I'll wait for a review from @whimsicallyson to make sure that's consistent, then we can merge/release tomorrow

whimsicallyson:
same-I'm getting about 6 sec loading time, but it _feels_ snappier and I wasn't able to get myself into a header-only state

znicholasbrown:
Yeah definitely something should improve on, whether through SSR or streamlining auth. For now we'll let this close #324

PR #355, opened 2020-10-20
Update the API connection menu to include helpful links for Server users

znicholasbrown:
PR Checklist:

## Describe this PR
Adds links to the docs, the Slack community, and Prefect Cloud for Server users that are having trouble connecting.

### DIFF
--- a/src/components/NavBar.vue
+++ b/src/components/NavBar.vue
@@ -1,10 +1,11 @@
 <script>
 import { mapActions, mapGetters, mapMutations } from 'vuex'
+import ExternalLink from '@/components/ExternalLink'
 import GlobalSearch from '@/components/GlobalSearchBar/GlobalSearch'
 import moment from '@/utils/moment'
 
 export default {
-  components: { GlobalSearch },
+  components: { ExternalLink, GlobalSearch },
   data() {
     return {
       connectionMenu: false,
@@ -238,7 +239,7 @@ export default {
           </v-icon>
         </v-btn>
       </template>
-      <v-card tile class="pa-0" max-width="320">
+      <v-card tile class="pa-0" max-width="500">
         <v-card-text class="pb-0">
           <div class="mb-2">
             <span v-if="connected">Connected</span>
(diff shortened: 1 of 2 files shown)


src/components/NavBar.vue
```diff
@@ -198,7 +199,6 @@ export default {
       v-model="connectionMenu"
       :close-on-content-click="false"
       offset-y
-      open-on-hover
```
zhen0, line 201: I actually like the open-on-hover - not sure I'd realize it was clickable and so might miss all the extra useful info.

src/components/NavBar.vue
```diff
@@ -241,14 +239,63 @@ export default {
...
+              <div>
+                Having trouble?
+                <span class="font-weight-medium">Don't panic! </span>
+              </div>
+              <div class="mt-4">
+                The <v-icon x-small>fab fa-slack</v-icon>&nbsp;
+                <ExternalLink
+                  href="https://join.slack.com/t/prefect-community/shared_invite/enQtODQ3MTA2MjI4OTgyLTliYjEyYzljNTc2OThlMDE4YmViYzk3NDU4Y2EzMWZiODM0NmU3NjM0NjIyNWY0MGIxOGQzODMxNDMxYWYyOTE"
+                  >Prefect Slack community</ExternalLink
+                >
```
zhen0, line 276: Should we also include a link to github discussions here?  There may already be a discussion around their issue?

zhen0 (changes requested):
Two small suggestions @znicholasbrown but otherwise looks like a great addition.

src/components/NavBar.vue
znicholasbrown: I'm thinking the consistency across the nav bar is better in this case, since all the other buttons on the top nav are on-click instead of on-hover. Perhaps a middle ground (but out of scope of this PR) would be to add tooltips to all the navbar buttons so it's easy to understand what they're meant for?

src/components/NavBar.vue
znicholasbrown: I added a link :)

zhen0 (approved):

PR #359, opened 2020-10-20
Accept invitations

### LINKED ISSUE #282: Update the design of the AcceptInvitation Page
**Why**

This is one of the first pages that many Cloud users will see but it was written without much focus on design. We should use it as a way to strengthen/show off the Prefect brand and also make the team name and details easier to read. 

**What**

* [ ] Updated design - The support page or onboard pages would make a good template for this one
* [ ] New tenant/team information is clearly displayed.  (Consider taking out user name as that's not always well chosen/edited on sign up)

**Notes**

zhen0:
PR Checklist:

## Describe this PR
- Updates the design of the accept invitation page to match other onboarding pages
- Closes #282 
- This page will only be shown to existing users who click on the emailed invitation link.  New users will be shown invitations (where relevant) in the updated name-team page. 
- We'll do a separate PR to update routing for onboarding (the accept route should be handled already by redirect route in auth nav guard)
- Also working on a separate PR to the docs to add information about teams and team switching in the UI

The page now looks like this: 

Screen Shot 2020-10-20 at 10 32 28 PM


src/pages/Onboard/AcceptInvitationPage.vue
```diff
@@ -0,0 +1,392 @@
...
+      } catch (e) {
+        this.mutationErrorMessage = e
+          .toString()
+          .split(':')
+          .pop()
+        this.loading = false
+        this.error = true
+      }
+    },
+    slash2Style(i) {
```
znicholasbrown, line 168: You shouldn't need the slash style computed props and methods if `accept` is a sub-route of `onboard` - you'll instead want to modify the classes on the parent route (I'd guess this is what's causing the moving slashes)

src/pages/Onboard/AcceptInvitationPage.vue
```diff
@@ -0,0 +1,392 @@
...
+      return (
+        this.error ||
+        !this.membershipInvitation ||
+        !this.membershipInvitation.user ||
+        this.mutationErrorMessage
+      )
+    },
+    errorMessage() {
+      if (this.membershipInvitation && !this.membershipInvitation.user)
+        return "The membership ID provided doesn't match this user. Are you in the right account? "
```
znicholasbrown, line 38:
```suggestion
        return "This invitation appears to be for someone else - did you log in to the correct account? "
```

src/pages/Onboard/AcceptInvitationPage.vue
```diff
@@ -0,0 +1,392 @@
...
+      if (this.error)
+        return 'It looks like there was a problem with your invitation.  Please check the details and your account and try again.'
+      return `We can't find your membership invitation. Have you already
```
znicholasbrown, line 41:
```suggestion
      return `We can't find this invitation... have you already
```

src/pages/Onboard/AcceptInvitationPage.vue
```diff
@@ -0,0 +1,392 @@
...
+    membershipInvitation: {
+      query: require('@/graphql/Tenant/membership-invitation.gql'),
+      variables() {
+        return { id: this.invitationId }
+      },
+      skip() {
+        return !this.invitationId
+      },
+      loadingKey: 'loadingKey',
+      pollInterval: 5000,
```
znicholasbrown, line 197: Do we need to poll for this?

src/pages/Onboard/AcceptInvitationPage.vue
```diff
@@ -0,0 +1,392 @@
...
+              Back to the dashboard!
+              <v-icon right>fas fa-rocket</v-icon>
+            </v-btn>
+          </div>
+        </v-col>
+      </v-row>
+      <v-row v-else align="center">
+        <v-col>
+          <div class="display-1 grey--text text--lighten-5">
+            You've been invited to join {{ teamName }}!
```
znicholasbrown, line 272: Let's make the team name stand out more, maybe add a semi-bold class to it and/or color the text?

src/pages/Onboard/AcceptInvitationPage.vue
```diff
@@ -0,0 +1,392 @@
...
+              Accept
+            </v-btn>
+
+            <v-btn
+              outlined
+              class="white--text"
+              :loading="deleting"
+              @click="decline"
+            >
+              No Thanks
```
znicholasbrown, line 296: Can we provide a confirmation when `No Thanks` is pressed? I can't think of many scenarios where this is the action the user intended to take and can see how it would be annoying if they accidentally pressed the wrong button.

src/pages/Onboard/AcceptInvitationPage.vue
```diff
@@ -0,0 +1,392 @@
...
+          <div class="ma-12 grey--text text--lighten-5 body-1 text--primary">
+            For more information about teams in Prefect Cloud, check out
+            <a
+              href="https://docs.prefect.io/orchestration/ui/team-settings.html"
+              target="_blank"
+              class="link-color"
+            >
+              our docs</a
+            >.
+          </div>
```
znicholasbrown, line 312: Nice 👍

src/pages/Onboard/AcceptInvitationPage.vue
```diff
@@ -0,0 +1,392 @@
...
+          return true
+        }
+      } catch (e) {
+        this.mutationErrorMessage = e
+          .toString()
+          .split(':')
+          .pop()
+        this.loading = false
+        this.error = true
+      }
```
znicholasbrown, line 146: I think this method is doing 2 different things - 1 it's returning the success of the mutation, the other it's setting an error but returning nothing. I think it'd make more sense to either 1) move the error logic into the `decline` method, keeping this method idempotent, or 2) moving the `if (declined)` block into the success statement here. I could also see a scenario where we don't need the first method at all, since it's sort of a wrapper for `declineInvitation`.

src/pages/Onboard/AcceptInvitationPage.vue
```diff
@@ -0,0 +1,392 @@
...
+        if (data?.accept_membership_invitation?.id) {
+          return true
+        }
+      } catch (e) {
+        this.mutationErrorMessage = e
+          .toString()
+          .split(':')
+          .pop()
+        this.loading = false
+        this.error = true
```
znicholasbrown, line 165: Same comment as above wrt idempotency

src/pages/Onboard/AcceptInvitationPage.vue
```diff
@@ -0,0 +1,392 @@
...
+      const invitationId = this.membershipInvitation.id
+      const declined = await this.declineInvitation(invitationId)
+      if (declined) {
+        this.toDashboard()
+      }
+      this.deleting = false
+    },
+    toDashboard() {
+      sessionStorage.removeItem('invitationId')
+      this.$router.push({ name: 'dashboard' })
```
znicholasbrown, line 126: I think this should include the tenant slug if possible.

src/pages/Onboard/AcceptInvitationPage.vue
```diff
@@ -0,0 +1,392 @@
...
+      const accepted = await this.acceptInvitation(invitationId)
+      if (accepted) {
+        sessionStorage.removeItem('invitationId')
+        await this.setCurrentTenant(tenant.slug)
+        this.$router.push({
+          name: 'dashboard',
+          params: { tenant: tenant.slug }
+        })
+      }
+      this.loading = false
```
znicholasbrown, line 113: This feels somewhat duplicative of the `toDashboard` method

znicholasbrown (changes requested):
This is looking really nice @zhen0! A few comments and also noticed this, which I don't think is intentional?
Oct-21-2020 11-16-13

zhen0:
Thanks for the comments @znicholasbrown - changes made and ready for another review.

zhen0:
Short addition the team settings docs that we link to from this page in [Prefect PR 3561](https://github.com/PrefectHQ/prefect/pull/3561)

src/pages/Onboard/Onboard-Page.vue
```diff
@@ -169,7 +176,8 @@ export default {
 
     <v-container class="position-absolute onboard-content pa-0" fluid>
       <transition name="fade" mode="out-in">
-        <router-view class="router-view" />
+        <acceptPage v-if="$route.name == 'accept'" />
```
znicholasbrown, line 179: Wouldn't it make more sense to add `accept` as a child route of `welcome`? Then this portion will be taken care of automatically and we won't have 2 routes pointing to the same component

src/router.js
```diff
@@ -36,7 +36,7 @@ export const routes = [
     path: '/accept',
     component: () =>
       import(
-        /* webpackChunkName: "accept" */ '@/pages/AcceptInvitationPage.vue'
+        /* webpackChunkName: "accept" */ '@/pages/Onboard/Onboard-Page.vue'
```
znicholasbrown, line 39: Same comment as above wrt child route of `welcome`

src/pages/Onboard/AcceptInvitationPage.vue
```diff
@@ -0,0 +1,221 @@
...
+                >
+              </v-card-actions>
+            </v-card>
+          </v-dialog>
+        </div>
+      </v-col>
+    </v-row>
+    <v-row align="center">
+      <v-col>
+        <div class="ma-12 grey--text text--lighten-5 body-1 text--primary">
```
znicholasbrown, line 202: Sorry, should have caught this before but can we make this text white? It's tough to see on the dark grey background.

znicholasbrown (changes requested):
2 small changes, otherwise tested and LGTM @zhen0

src/pages/Onboard/Onboard-Page.vue
zhen0: Yes we can.  I was going to do that as part of a separate PR though as the invitation email currently sends to _cloud.prefect.io/accept_ so it would involve changing the routing either as part of a navguard or watch on app or as a PR to Cloud.

znicholasbrown (approved):
LGTM @zhen0 😄

Issue #329, opened 2020-10-14
Have the Project's name as Prefix when Creating a Cloud Hook


raphael-riel:
**Why**
I have flows with the same name under different projects (Prod, Test, Dev, etc.)
I'm unable to properly identify which flow is which when I attempt to create a Cloud Hook.

**What**
* [ ] Prefix Flow's name with their project's name (Or on roll-over maybe, up to you!)

**Notes**
![image](https://user-images.githubusercontent.com/474697/96011337-cd34fe80-0e10-11eb-94da-4ef8a7080f50.png)

PR #363, opened 2020-10-21
Cloud Hook Prefix

ThatGalNatalie:
PR Checklist:

## Describe this PR
Adds the project name as a prefix when creating a Cloud Hook. Before, flows with the same name under different projects (Prod, Test, Dev, etc.) were hard to identify which flow was which when creating a Cloud Hook.

### DIFF
--- a/src/components/CloudHookForm.vue
+++ b/src/components/CloudHookForm.vue
@@ -165,7 +165,11 @@ export default {
       return allHooks.filter(
         t => (t.requiresCloud && this.isCloud) || !t.requiresCloud
       )
-    }
+    },
+    projectNamePrefix() {
+      const prefixedName = item => item.name + " ("+item.project.name+")"
+      return prefixedName
+    },
   },
   watch: {
     loading(val) {
@@ -406,7 +410,7 @@ export default {
                   v-model="versionGroupIdSelect"
                   data-cy="flow-list"
                   item-value="version_group_id"
-                  item-text="name"
+                  :item-text="projectNamePrefix"
                   label="Flow (Version Group)"
                   class="headline overflow"
                   :items="flows"
--- a/src/components/CloudHookForm.vue
+++ b/src/components/CloudHookForm.vue
@@ -407,6 +407,7 @@ export default {
               <v-col cols="12" md="8" class="pb-0">
                 <v-autocomplete
                   v-if="canEdit && !versionGroupIdProp"
+                  id="item"
                   v-model="versionGroupIdSelect"
                   data-cy="flow-list"
                   item-value="version_group_id"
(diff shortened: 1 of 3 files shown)


src/graphql/TeamSettings/flows.gql
```diff
@@ -4,5 +4,8 @@ query Flows {
     archived
     name
     version_group_id
+    project {
```
znicholasbrown, line 7: Could you add `id` in the project join here? Vue-Apollo doesn't do well without that field and can return really weird results in some caching scenarios.

src/components/CloudHookForm.vue
```diff
@@ -403,15 +407,24 @@ export default {
               <v-col cols="12" md="8" class="pb-0">
                 <v-autocomplete
                   v-if="canEdit && !versionGroupIdProp"
+                  id="item"
                   v-model="versionGroupIdSelect"
                   data-cy="flow-list"
                   item-value="version_group_id"
-                  item-text="name"
+                  :item-text="projectNamePrefix"
```
znicholasbrown, line 414: This is very clever!

src/components/CloudHookForm.vue
```diff
@@ -165,6 +165,10 @@ export default {
       return allHooks.filter(
         t => (t.requiresCloud && this.isCloud) || !t.requiresCloud
       )
+    },
+    projectNamePrefix() {
```
znicholasbrown, line 169: Could this be moved to `methods` instead? I'm not sure it needs to be a computed prop.

src/components/CloudHookForm.vue
```diff
@@ -165,6 +165,10 @@ export default {
...
+      const prefixedName = item => item.name + ' (' + item.project.name + ')'
```
znicholasbrown, line 170:
```suggestion
      const prefixedName = item => `item.name (${item.project.name})`
```
I'd suggest interpolating this string instead, it's a little easier to maintain. You can also directly return this function instead of allocating a variable first

znicholasbrown (changes requested):
A couple of small suggestions but looks great otherwise @ThatGalNatalie !

src/components/CloudHookForm.vue
ThatGalNatalie: Thanks!

src/components/CloudHookForm.vue
ThatGalNatalie: Yeah, I'm still learning more about the difference between methods vs computed prop but I thought it might be better because it only runs again if a reactive dependency changes but in this case, I don't think it makes too much of a difference. I'll move it to methods.

ThatGalNatalie:
just added all the suggestions and I think it's good to go now @znicholasbrown

znicholasbrown (approved):
Gj @ThatGalNatalie 

Resolves: #329
cc @raphael-riel

PR #373, opened 2020-10-23
Late count indicator

whimsicallyson:
PR Checklist:

- [X] add a short description of what's changed to the top of the `CHANGELOG.md`

Adds a count indicator for upcoming and late runs to the tile. Note that this currently shows the exact number of each, as I didn't notice any issues with overflowing the container for 2 or 3 digit numbers and didn't expect larger. Happy to truncate large counts if we think that's better.

Screen Shot 2020-10-23 at 11 08 29 AM

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -4,6 +4,7 @@
 
 ### Features and Improvements
 
+- Add indicator of number of upcoming or late runs to the dashboard card title = [#348](https://github.com/PrefectHQ/ui/issues/348)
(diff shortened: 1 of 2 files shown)


znicholasbrown:
This looks good @whimsicallyson - could we hide the counts if there are 0 or if you're on that tab?

znicholasbrown:
This LGTM @whimsicallyson - I'll approve after that conflict is resolved 👍

znicholasbrown (approved):

PR #382, opened 2020-10-28
update no agents warning 

### LINKED ISSUE #215: Make it more obvious that a user needs to start an agent

zhen0:
PR Checklist:

## Describe this PR
- Changes the no agents tile color from 'warning' orange to 'error' red and updates the messages to say flows will not run without agents.  
- Also adds a link to the agents docs. 
- Together with the work on flow affinity labels, I think this closes #215

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -15,7 +15,6 @@
 ### Features and Improvements
 
 - Add ability to accept team invites on new user creation, go directly to invited team - [#322](https://github.com/PrefectHQ/ui/issues/322)
-- Add indicator of number of upcoming or late runs to the dashboard card title = [#348](https://github.com/PrefectHQ/ui/issues/348)
 - Update the agents tile on the dashboard to show flows can not run without agents [#382](https://github.com/PrefectHQ/ui/pull/382)
 - Add indicator of number of upcoming or late runs to the dashboard card title - [#348](https://github.com/PrefectHQ/ui/issues/348)
 
--- a/Changelog.md
+++ b/Changelog.md
@@ -5,6 +5,7 @@
 ### Features and Improvements
 
 - Add indicator of number of upcoming or late runs to the dashboard card title = [#348](https://github.com/PrefectHQ/ui/issues/348)
+- Update the agents tile on the dashboard to show flows can not run without agents [#382](https://github.com/PrefectHQ/ui/pull/382)
 
 ### Bugfixes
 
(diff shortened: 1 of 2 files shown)


madkinsz:
I noticed the first link creates a new tab but the other two navigate within the same tab--this makes some sense since the second two are within the same app but may be worth creating a new tab for all 3 anyway.

This may be a useful message to display if all of your agents are not running as well? I only had one, failed agent and a pointer to the agent docs would have been helpful yet.

zhen0:
> I noticed the first link creates a new tab but the other two navigate within the same tab--this makes some sense since the second two are within the same app but may be worth creating a new tab for all 3 anyway.
> 
> This may be a useful message to display if all of your agents are not running as well? I only had one, failed agent and a pointer to the agent docs would have been helpful yet.

I only added one link here?

Good point on displaying the link to the docs for any error - I've added that.

znicholasbrown:
Screen Shot 2020-10-29 at 11 35 30 AM
This looks pretty out of place as-is, could we move it to be a continuation of the line above?

madkinsz:
> > I noticed the first link creates a new tab but the other two navigate within the same tab--this makes some sense since the second two are within the same app but may be worth creating a new tab for all 3 anyway.
> > This may be a useful message to display if all of your agents are not running as well? I only had one, failed agent and a pointer to the agent docs would have been helpful yet.
> 
> I only added one link here?

Oops sorry! I should have taken a look at the code. I was looking at 

Screen Shot 2020-10-29 at 11 04 22 AM

zhen0:
Screen Shot 2020-10-29 at 12 18 15 PM
 
Adding as a continuation of the line above created a big-ish text block that I find harder to read/parse.  As a compromise I've added as part of the list item so it's in-line but a separate paragraph.

znicholasbrown (approved):
LGTM 👍

PR #390, opened 2020-10-29
md parser 

ThatGalNatalie:
PR Checklist:

## Describe this PR

- Added a md parser ([remark](https://www.npmjs.com/package/remark)) for the upcoming artifacts/tutorials
- Created an "artifacts" tab in the flow run and task run


zhen0 (commented):
This is looking good @ThatGalNatalie - markdown works when visiting the tutorial test page.  

A few thoughts and comments: 
- The coming soon badge isn't disabled but I think that's a Vuetify oddity we can look into together. (But I don't think we can merge until fixed.)
- Should/could we add a bit of formatting (maybe just some margin or padding around the edges) to the markdown page and tabs? - That might be better as a future PR but putting here for discussion. 
- Artifact or Artifacts for the header?  I think I'd pluralize but not certain if any flow run would have more than one artifact?

znicholasbrown:
We'll definitely want to pluralize artifacts, and I think we should use a different icon. Perhaps `fas fa-fingerprint`?:
Screen Shot 2020-11-11 at 1 32 41 PM

Agreed with @zhen0's review but we can probably save the markdown page formatting for later since the point of this is to get the parser into the codebase so we can use it for artifacts.

zhen0:
Agreed on saving formatting for later - love that fingerprint icon!

ThatGalNatalie:
Fixed the disabled issue and changed the icon and name. I think it's good to go now unless there was something else 
@znicholasbrown @zhen0

znicholasbrown (approved):
LGTM @ThatGalNatalie !

PR #395, opened 2020-11-02
Make agents and upcoming tiles refresh when switch backend

### LINKED ISSUE #347: BUG: Late and upcoming runs tiles on the Flow and Dashboard pages don't correctly refresh data
**Why**

When switching back to the dashboard or flow pages after some time away, the data sitting on the component is stale. Polling to update the data starts immediately but sometimes takes as much as 5 seconds to return, causing the tiles to show "Late" runs incorrectly. 

**What**

* [ ] the upcoming/late runs tiles correctly destroy their data on destroy so they don't show incorrect data when returning to the tile after some time away

zhen0:
PR Checklist:

## Describe this PR
- Makes sure the upcoming runs and agents tiles refresh when switch backends and do not use cached data with a server-less backend.  (I thought this was happening by default but adding fetching policy of 'no-cache' seemed to make the important difference). 
- Also updates the upcoming tile tab - I was seeing the upcoming runs tile stay on the late tab even when there were no late runs showing and there upcoming runs to display.  
- It also deals with a few small console error messages that arose when clearing late runs. 
- Should close #347


src/App.vue
```diff
@@ -251,14 +252,12 @@ export default {
...
-          (this.isServer && !this.connected)
-        )
-      },
-      update: data => data.agent
+      //Without this, server UI with no actual server shows results
+      fetchPolicy: 'no-cache',
+      update({ agent }) {
+        if (!agent) return null
+        return agent
+      }
```
znicholasbrown, line 260: I'm not sure I understand why this change was made, unless we need `null` specifically, in which case we could do:
```js
update: data => data.agent ?? null
```

znicholasbrown (commented):
This looks good @zhen0, just 1 comment/question.

src/App.vue
zhen0: Yup wanted to be sure we had null.  But I like your suggestion better - I'll update.

znicholasbrown (approved):

znicholasbrown (approved):

Issue #366, opened 2020-10-22
BUG: Scroll bars CSS issue


kfirstri:
**Bug Description**
There are empty scroll bars on all "card" elements in the UI

Dashboard - 
![image](https://user-images.githubusercontent.com/3606980/96846818-101f5380-145b-11eb-9653-acb94072f920.png)

Task View -
![image (6)](https://user-images.githubusercontent.com/3606980/96846827-131a4400-145b-11eb-94a5-e08b5630360c.png)

Flow View -
![image (5)](https://user-images.githubusercontent.com/3606980/96846830-131a4400-145b-11eb-870d-015dfa0d5f4f.png)

Agent Card -
![image (4)](https://user-images.githubusercontent.com/3606980/96846832-13b2da80-145b-11eb-931d-f5bfd3257a2e.png)

Disabling the CSS property
![image (7)](https://user-images.githubusercontent.com/3606980/96846825-11e91700-145b-11eb-9aed-7e3a105e8d96.png)

if I disable this css prop - 
```css
.overflow-y-scroll {
    overflow-y: scroll;
}
```
Then the scroll lines disappear but then the content overflows it's container

**Steps To Reproduce**
1. Install prefect with pip `pip install prefect` (latest version
2. run the server
3. go to localhost:8080

**Browsers Tested:**
* [X] Chrome - Version 86.0.4240.80 (Official Build) (x86_64)
* [X] Firefox
* [X] Safari
* [ ] Edge
* [ ] IE

whimsicallyson:
hi @kfirstri! I'm taking a look at this bug, and I'm able to reproduce it while my colleague isn't. We think the difference is in the MacOS we're running--out of curiosity, what OS are you on?

kfirstri:
Hi @whimsicallyson :)
Giving you lots of specs just to make sure - 
Model Name:	        MacBook Pro
Model Identifier:	MacBookPro14,3
Processor Name:	Quad-Core Intel Core i7
System Version:	macOS 10.15.7 (19H2)
Kernel Version:	Darwin 19.6.0

PR #399, opened 2020-11-04
Scrollbar bug

whimsicallyson:
PR Checklist:

- [X] add a short description of what's changed to the top of the `CHANGELOG.md`

## Describe this PR
User reported, and I also saw, unnecessary scrollbars showing up on many/most tiles on the assorted dashboards, even when there was not information overflowing the container. Strangely, @zhen0 wasn't seeing this on her computer, even with the same version of Chrome? Changed `overflow-y: scroll` to `overflow-y: auto` in many places to fix this, and removed the (now unused) `overflow-y-scroll` global class.

Note: I also made some changes to the overall layout of the Details tile, as it was using a `div` instead of the `v-list` the other tiles were using and causing different paddings.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -8,7 +8,7 @@
 
 ### Bugfixes
 
-- None
+- Fix issue where scrollbars show on all tiles even when not needed - [#366](https://github.com/PrefectHQ/ui/issues/366)
 
 ## 2020-10-29a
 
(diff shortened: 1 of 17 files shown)


znicholasbrown (approved):
I can't reproduce the original issue (still) but I think this is a good tech-debt reduction regardless. Once this is in prod we can have the OP in https://github.com/PrefectHQ/ui/issues/366 test to see if it fixes their issue and close that ticket if so.

PR #400, opened 2020-11-05
Add more info to flow Run Config details section

jcrist:
Previously we only displayed the `type`, we now display a select subset of the fields for each (known) type of `RunConfig`. Only non-null fields are displayed, and none of the values should be terribly large, so for most users things should display nicely.

PR Checklist:

### DIFF
--- a/src/pages/Flow/Details-Tile.vue
+++ b/src/pages/Flow/Details-Tile.vue
@@ -70,6 +70,28 @@ export default {
           return obj
         }, {})
     },
+    runConfigDisplayFields() {
+      if (!this.flow.run_config || !this.flow.run_config.type) return []
+
+      let fields = ['type']
+      if (this.flow.run_config.type == 'LocalRun') {
+        fields.push('working_dir')
+      } else if (this.flow.run_config.type == 'DockerRun') {
+        fields.push('image')
+      } else if (this.flow.run_config.type == 'KubernetesRun') {
+        fields.push(
+          'image',
+          'job_template_path',
+          'cpu_request',
+          'cpu_limit',
+          'memory_request',
+          'memory_limit'
+        )
+      } else if (this.flow.run_config.type == 'ECSRun') {
+        fields.push('image', 'task_definition_path', 'cpu', 'memory')
+      }
+      return fields.filter(field => this.flow.run_config[field] != null)
+    },
     flows() {
       if (!this.flow.storage || !this.flow.storage.flows) return null
       return this.flow.storage.flows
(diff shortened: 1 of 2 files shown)


jcrist:
Example:


Screen Shot 2020-11-05 at 4 15 18 PM

src/pages/Flow/Details-Tile.vue
```diff
@@ -70,6 +70,28 @@ export default {
...
+    runConfigDisplayFields() {
+      if (!this.flow.run_config || !this.flow.run_config.type) return []
+
+      let fields = ['type']
+      if (this.flow.run_config.type == 'LocalRun') {
+        fields.push('working_dir')
+      } else if (this.flow.run_config.type == 'DockerRun') {
+        fields.push('image')
+      } else if (this.flow.run_config.type == 'KubernetesRun') {
+        fields.push(
```
zhen0, line 82: Do we need to hard code these? I wonder if there might be a way we could pull them dynamically instead?  Maybe have the runConfigFields be the keys of the run_config object (dict) but with type, labels and env filtered out?  Otherwise, we'll need to manually update if/when any changes are made in core.  Could be good reasons - just want to check.

zhen0 (commented):
Nice changes but one question @jcrist

src/pages/Flow/Details-Tile.vue
jcrist: Since some fields might be large or undesired, I opted for a hardcoded allow-list for each type. The set of fields are unlikely to change rapidly in core, so I don't think hardcoding it in the UI is a big issue, and it'll be straightforward to add new fields later if needed. It also lets us display fields in nicer logical order, rather than random/alphabetical.

zhen0: Sounds good to me!

zhen0:
I'll let you do a final review @znicholasbrown

src/pages/Flow/Details-Tile.vue
```diff
@@ -70,6 +70,28 @@ export default {
...
+          'job_template_path',
+          'cpu_request',
+          'cpu_limit',
+          'memory_request',
+          'memory_limit'
+        )
+      } else if (this.flow.run_config.type == 'ECSRun') {
+        fields.push('image', 'task_definition_path', 'cpu', 'memory')
+      }
+      return fields.filter(field => this.flow.run_config[field] != null)
```
znicholasbrown, line 93:
```suggestion
      return fields.filter(field => field in this.flow.run_config)
```
Let's use this instead because there are a lot of edge cases when using `!= null`; this will check that the property exists. If we specifically need to filter out `null` (rather than `undefined`), we could instead us `!== null`

znicholasbrown (changes requested):
Just one comment @jcrist, otherwise LGTM

src/pages/Flow/Details-Tile.vue
jcrist: I'm trying to filter out fields that are missing or are explicitly set as `null` (but do want to keep other false-y values like `""` or `0`). Does this do that? I am not a JS dev, pulled this from this [stackoverflow answer](https://stackoverflow.com/a/15992131), which seemed to imply that `undefined == null` is true.

znicholasbrown (approved):
Discussed comments offline and this looks good 👍

Issue #194, opened 2020-09-03
Add a "Your Teams" section to the user page


zhen0:
**Why**
It would be useful to see information about all the teams that I am a member of and my role in those teams.  Ideally I could leave a team from that page. 

**What**
* [x] New "Your Teams" page/ section of the User Page
The table shows: 
* [x] the team (i.e. tenant) names, 
* [x] team (tenant) URL/Slug (which you see when you're on the dashboard so it's how I identify tenants)
* [x] user's role in the team
* [x] A link to the team (or says current if they're in that tenant)
* [x] A leave team button that removes them from that tenant 

**Notes**
Example table:

**Your teams**

|__Name__|  ___URL___    |                __Role__             | | |
|----|-------------------|---------------------|-------|--|
|Team 1   |   teamone        |     Administrator  |  [View] |  [Leave] |
|Team 2   |    teamtwo     |       User               |  Current | [Leave] |
|Team 3   |   teamthree     |  Administrator  |  [View]  | [Leave] |

PR #403, opened 2020-11-06
Create a "Your Teams" page in the User prefs

whimsicallyson:
PR Checklist:

## Describe this PR

Implements the enhancement in #194 to make a page with a table of all the teams/tenants you're a part of, with the ability to switch between them (switching will take you to the dashboard) or leave teams you don't need to be in. A few comments inline...


src/pages/UserSettings/Teams.vue
```diff
@@ -0,0 +1,171 @@
...
+    },
+    async removeUser(removalTenant) {
+      this.isRemovingUser = true
+      const initialTenant = this.tenant
+
+      let membershipsObject = {}
+      for (const membership of this.memberships) {
+        membershipsObject[membership.tenant.id] = membership.id
+      }
+      const membershipId = membershipsObject[removalTenant.id]
```
whimsicallyson, line 58: there's probably a better way to do this? but I wanted to get just the relevant information for identifying and switching tenants into one spot

src/pages/UserSettings/Teams.vue
```diff
@@ -0,0 +1,171 @@
...
+
+      if (initialTenant.id !== removalTenant.id) {
+        await this.setCurrentTenant(removalTenant.slug)
+      }
```
whimsicallyson, line 62: You can only leave the tenant you're currently logged into. Cloud issue #2827 would make this easier, but this will work for now.

src/pages/UserSettings/Teams.vue
```diff
@@ -0,0 +1,171 @@
...
+        )
+      }
+
+      if (initialTenant.id !== removalTenant.id) {
+        await this.setCurrentTenant(initialTenant.slug)
+      } else if (this.tenants[0].id === removalTenant.id) {
+        await this.setCurrentTenant(this.tenants[1].slug)
+      } else {
+        await this.setCurrentTenant(this.tenants[0].slug)
+      }
```
whimsicallyson, line 87: I don't love this, but needed to figure out a way to switch to a tenant you're still a part of. Happy to hear suggestions!

src/pages/UserSettings/Teams.vue
```diff
@@ -0,0 +1,171 @@
...
+  }
+}
+</script>
+
+<template>
+  <ManagementLayout>
+    <template #title>Your Teams</template>
+
+    <template #subtitle>
+      See your Prefect teams and roles, switch between them, and leave teams
```
whimsicallyson, line 118: Could use wordsmithing.

src/pages/UserSettings/Teams.vue
```diff
@@ -0,0 +1,171 @@
...
+      headers: [
+        { text: 'Name', value: 'name' },
+        { text: 'URL', value: 'slug' },
+        { text: 'Role', value: 'role' },
+        { text: 'Actions', value: 'id' }
+      ],
+      isRemovingUser: false,
+      removeTenant: null
+    }
+  },
```
zhen0, line 25: Could we refetch user on mount?  (I think the vuex action is getUser?) I accepted an invitation but couldn't see that team until I refreshed my page.

zhen0: Screen Shot 2020-11-09 at 5 53 06 PM

zhen0: 4 teams in sideNav but only 3 in the teams page.

src/pages/UserSettings/Teams.vue
```diff
@@ -0,0 +1,171 @@
...
+            `Are you sure you want to remove yourself from ${removeTenant.name}?`
+          "
+          :dialog-props="{ 'max-width': '600' }"
+          :disabled="isRemovingUser"
+          :loading="isRemovingUser"
+          @confirm="removeUser(removeTenant)"
+        >
+          <div class="red--text"
+            >Are you sure you want to remove yourself from
+            {{ removeTenant.name }}? You'll no longer be able to access your
```
zhen0, line 164: Maybe switch this to something like "run data associated with {{removeTenant.name}}"?  They should still be able to access their flow code etc.

zhen0 (commented):
This worked really nicely for me @whimsicallyson .  I was able to switch between my tenants and leave one with no problems - really good to have the warning first too.  One slight issue that I think is more from Cloud than this PR but it affects it.  Then a few small thoughts on design and wording.

src/pages/UserSettings/Teams.vue
znicholasbrown:
```suggestion
      See your Prefect teams and roles, switch between them, and leave teams
```
View and manage your team memberships

src/pages/UserSettings/Teams.vue
```diff
@@ -0,0 +1,171 @@
...
+      :items="tenants"
+      :footer-props="{
+        showFirstLastPage: true,
+        firstIcon: 'first_page',
+        lastIcon: 'last_page',
+        prevIcon: 'keyboard_arrow_left',
+        nextIcon: 'keyboard_arrow_right'
+      }"
+    >
+      <template #item.role="{item}">
```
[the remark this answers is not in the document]
znicholasbrown: Agreed with @zhen0 - can probably use the icon buttons we have elsewhere, which will help with spacing and visual consistency.

src/pages/UserSettings/Teams.vue
```diff
@@ -0,0 +1,171 @@
...
+    <template #title>Your Teams</template>
+
+    <template #subtitle>
+      See your Prefect teams and roles, switch between them, and leave teams
+    </template>
+
+    <v-data-table
+      fixed-header
+      class="elevation-2 rounded-none truncate-table"
+      :headers="headers"
```
znicholasbrown, line 124:
```suggestion
      :headers="headers"
      :header-props="{ 'sort-icon': 'arrow_drop_up' }"
```

Adding this line will make sure the sort arrows are consistent with other tables.

znicholasbrown (changes requested):
This works well @whimsicallyson - I left a few comments and agree with @zhen0's review. In addition, I noticed that the table is showing a tenant to which I've been invited but haven't accepted or declined but isn't distinguishing between it and teams I'm already a member of. It'd be good for some visual indicator that it's a pending team (and maybe allow accepting/declining from this table as well? Could re-use the component from the notifications or the tenant switcher in the sidenav).

src/pages/UserSettings/Teams.vue
whimsicallyson: huh, this is the same wording used if you remove yourself from a team on the members page. If we want to update it, we should probably change it both places...

src/pages/UserSettings/Teams.vue
```diff
@@ -0,0 +1,171 @@
...
+      removeTenant: null
+    }
+  },
+  computed: {
+    ...mapGetters('user', ['memberships']),
+    ...mapGetters('tenant', ['tenant', 'tenants'])
+  },
+  methods: {
+    ...mapActions('tenant', ['getTenants', 'setCurrentTenant']),
+    role(item) {
```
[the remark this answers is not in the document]
whimsicallyson: opened https://github.com/PrefectHQ/cloud/issues/2831, feel free to add context/clarification if needed!

src/pages/UserSettings/Teams.vue
whimsicallyson: so I've added a `getUser` and `getTenants` after accepting or declining a team invite, but its not updating the way I'd expect--specifically, I'd like to see the action icons get replaced with the relevant ones after joining a team, but the old ones just go away and are replaced with a blank space :(

src/pages/UserSettings/Teams.vue
whimsicallyson: updated in both locations!

src/pages/UserSettings/Teams.vue
znicholasbrown: That's most likely because of https://github.com/PrefectHQ/ui/pull/403/files#diff-ceb8a586ffaa424f678e44816b6a17d8aa390c65ad1728a6cb1360570c4bbd3eR227, https://github.com/PrefectHQ/ui/pull/403/files#diff-ceb8a586ffaa424f678e44816b6a17d8aa390c65ad1728a6cb1360570c4bbd3eR242, and https://github.com/PrefectHQ/ui/pull/403/files#diff-ceb8a586ffaa424f678e44816b6a17d8aa390c65ad1728a6cb1360570c4bbd3eR257. You're referencing a computed prop by calling its getter method instead of its reactive prop - try `currentMemberships.includes(item.id)` instead.

whimsicallyson: still not sure its working...

zhen0: I think something like `async mounted() {
    await this.getTenants()
  },` should work?

zhen0: https://blog.logrocket.com/introduction-to-vue-lifecycle-hooks/

zhen0 (commented):
This is looking really nice - the new icons work well and alignment is much better.  I've added a suggestion for getTenants() on mount that I _think_ will help with making sure all tenants show up.  

Adding gettenants still doesn't add/show an invitation if the invitation arrives while a user is on the teams page - one way round that would be to add a tenant query instead of using vuex.  But I'm not sure it's a major issue so not something I think we **need** to do here - but maybe @znicholasbrown has an opinion on that?

src/components/AcceptConfirmInputRow.vue
```diff
@@ -32,22 +36,31 @@ export default {
...
-        <v-btn color="error" text @click="confirm = 'decline'">
-          <v-icon>
-            close
-          </v-icon>
-        </v-btn>
+        <v-tooltip :disabled="!tooltips" bottom
+          ><template #activator="{on}"
+            ><v-btn color="error" text v-on="on" @click="confirm = 'decline'">
+              <v-icon>
+                close
```
whimsicallyson, line 59: this eye is so much heavier than the checkmark, I'm tempted to either shrink it or find a different icon. Anyone else feel strongly?

src/pages/UserSettings/Teams.vue
```diff
@@ -243,37 +243,38 @@ export default {
...
+            ><template #activator="{on}"
+              ><v-btn
+                v-if="item.tenant.id !== tenant.id && item.member"
+                text
+                small
+                style="margin-right: 14px;"
+                color="primary"
+                v-on="on"
+                @click="handleSwitchTenant(item)"
+                ><v-icon>fas fa-eye</v-icon></v-btn
```
whimsicallyson, line 257: this eye is so much heavier than the checkmark, I'm tempted to either shrink it or find a different icon. Anyone else feel strongly?

zhen0: I think font awesome icons tend to be larger for some reason.  You could add a 'small' prop to the v-icon?

znicholasbrown: How about these? (`swap_horiz` and `meeting_room`)
Screen Shot 2020-11-11 at 4 31 59 PM

whimsicallyson: I like swap_horiz, my feeling is that declining an invite and leaving a team are close enough to the same thing that they can have the same icon.

znicholasbrown (approved):
LGTM @whimsicallyson !

PR #411, opened 2020-11-09
fetch task runs for mark as dialog

zhen0:
PR Checklist:

## Describe this PR
- Fixes a bug on the set state button on the flow run page where the option to include task run states was not working.

### DIFF
--- a/src/graphql/FlowRun/task-run-ids.gql
+++ b/src/graphql/FlowRun/task-run-ids.gql
@@ -1,0 +1,9 @@
+query taskRunIds($flowRunId: uuid!) {
+  flow_run_by_pk(id: $flowRunId) {
+      id
+    task_runs{
+    version
+      id
+    }
+  }
+}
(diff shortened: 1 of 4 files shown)


znicholasbrown (changes requested):
This will most likely fail for large mapped pipelines - I'm not sure how we want to address that UX. Maybe we should filter out task runs with `> -1` map index? Other than that, just one issue to fix.

zhen0:
Actually - not ready for review - just seen something else that needs updating...

src/graphql/FlowRun/task-run-ids.gql
```diff
@@ -0,0 +1,7 @@
+query taskRunIds($flowRunId: uuid!, $parentMapIndex: Int, $childMapIndex: Int) {
+    task_run (where: {_and: [{flow_run_id: {_eq: $flowRunId}}, {map_index: {_eq: $parentMapIndex}}, {map_index: { _gte: $childMapIndex}}]}){
+    version
+      id
+      map_index
+  }
+}
```
znicholasbrown, line 7:
```suggestion
query taskRunIds($flowRunId: uuid!, $parentMapIndex: Int, $childMapIndex: Int) {
  task_run(
    where: {
      _and: [
        { flow_run_id: { _eq: $flowRunId } }
        { map_index: { _eq: $parentMapIndex } }
        { map_index: { _gte: $childMapIndex } }
      ]
    }
  ) {
    version
    id
    map_index
  }
}

```
Just to clean this file up a bit

znicholasbrown (changes requested):
This is working great @zhen0, just 1 suggestion to clean up the task run ids query file.

zhen0:
Thanks @znicholasbrown - made that clean up.  (This whole feature needs a re-factor at some point!)

znicholasbrown (approved):

PR #417, opened 2020-11-10
Accept invitation

### LINKED ISSUE #381: Consider component-izing the logic for the list of pending team invitations
Currently the same logic is repeated in the sidenav, name team and notifications (E.g. see lines 293-323 in the sidenav component)
We should look at whether there is a place we can consolidate the query and the filter.  This could be in a Vuex store or possibly we could create a more efficient query.

[ ] Look into whether a more efficient apollo query is possible
[ ] If not, check whether we could move the query and the filtering to Vuex
[ ] Investigate whether we could create a component to deal with all the pending invitations logic in one place. 
[ ] If a new component makes sense, create one!  (Or note why not.)

zhen0:
PR Checklist:

## Describe this PR
- Removes the unnecessary logic in the pending invitations query in the sidenav and name team pages
- Closes #381

### DIFF
--- a/src/pages/Onboard/NameTeam.vue
+++ b/src/pages/Onboard/NameTeam.vue
@@ -17,8 +17,7 @@ export default {
       currentInvitation: null,
       declining: false,
       dialog: false,
-      redirectTenant: null,
-      pendingInvitations: [],
+      redirectTenant: null
 
       // Reveal animation bools
       revealNote: false,
--- a/src/pages/Onboard/NameTeam.vue
+++ b/src/pages/Onboard/NameTeam.vue
@@ -17,7 +17,7 @@ export default {
       currentInvitation: null,
       declining: false,
       dialog: false,
-      redirectTenant: null
+      redirectTenant: null,
 
       // Reveal animation bools
       revealNote: false,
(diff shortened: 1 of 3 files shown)


whimsicallyson:
just tested with the teams page stuff I'm working on and it all acted the same, LGTM!

src/pages/Onboard/NameTeam.vue
```diff
@@ -11,14 +11,14 @@ export default {
...
       height: '0px',
-
       loading: 0,
       accepting: false,
       activeInvite: null,
       currentInvitation: null,
       declining: false,
       dialog: false,
       redirectTenant: null,
+      pendingInvitations: [],
```
znicholasbrown, line 21: We shouldn't need this since `pendingInvitations` will be moved to the top of the component by apollo anyway (same with that in the sidenav, now that the extra logic is removed)

znicholasbrown (dismissed):
One comment otherwise this looks good @zhen0 👍

src/pages/Onboard/NameTeam.vue
zhen0: That's what I thought... but then I needed to add in pendingInvitations && pendingInvitations.length wherever we v-if it....

znicholasbrown: Ah yea, then you can modify this line like this:
```
 update: data => data.pendingInvitations ?? []
```

znicholasbrown (approved):
LGTM 👍

Issue #422, opened 2020-11-10
Enable flow run re-scheduling from the calendar view page (possibly drag and drop?)


zhen0:
**Why**

**What**

* [ ] 

**Notes**

Issue #421, opened 2020-11-10
Add create schedule functionality to calendar view page


zhen0:
**Why**

**What**

* [ ] 

**Notes**

Issue #420, opened 2020-11-10
Add create flow run functionality to calendar view page


zhen0:
**Why**

**What**

* [ ] 

**Notes**

Issue #419, opened 2020-11-10
Filter calendar view page by project/agent/state/flow


zhen0:
**Why**

**What**

* [ ] 

**Notes**

PR #433, opened 2020-11-12
Calendar view

zhen0:
PR Checklist:


zhen0:
First part of addressing #418 which in turn is the first part of #197

zhen0:
Commit history got mixed here.  Will fix on Monday.  Latest commits add a date picker that selects date for day view and filters by day start and end - but still get flow query time outs.  Still to do:
* [ ] Add running flows
* [ ]  Move to sidenav not dashboard
* [ ] Add scheduled flows
* [ ] Add grouping for flows with a lot of runs
* [ ] Add loading states
* [ ] Improve/update colors, fonts, margins etc

zhen0:
Much updated now shows just one flow at a time and uses the interval slot to give more granular control over which events we show and how we show them.  Next steps:
* [ ] Improve loading time and handling
* [ ] Enable selection of multiple flows
* [ ] Make sure choosing a scheduled flow updates the calendar
* [ ] Add distinct on to scheduled runs query to make sure only id per flow
* [ ] Improve info, styling and handling in popover
* [ ] Improve styling of events 
* [ ] Improve handling of where there's still too many runs for a time period

zhen0:
* [ ] Improve info, styling and handling in popover - add chart with task run info - poss doughnut
* [ ] Margins and padding
* [ ] Make sure initial flow run is marked as selected flow
* [ ] Add flow name header for four day view
* [ ] Hour view
* [ ] Consider dynamic interval length

zhen0:
Added chart with task run info but wonder if that might be better being placed in a separate PR and focussing on neatening calendar view

src/graphql/Calendar/flow-run-task-runs.gql
```diff
@@ -0,0 +1,77 @@
+query RingChart(
```
whimsicallyson, line 1: any reason this query name doesn't match up with the file name?

whimsicallyson:
one thing I'm noticing while playing with the preview is that this snackbar shows/hides a lot, I'm tempted to suggest having it persistent and remember that its been dismissed -- the first time I saw it, I clicked the dismiss button and was surprised to see it again when I clicked over to another flow
Screen Shot 2020-12-04 at 4 41 46 PM

src/graphql/Calendar/flow-run-task-runs.gql
zhen0: Actually - this file will be coming out in this PR.  But good point/question about making them match.

zhen0:
Ideas for future iterations (I'll update on #418 - the calendar view ticket once this is merged):
 - Full week and month view
 - Filter by state, project, agent etc  - #419
 - Add create flow run functionality to the calendar view - #420
- Add create schedule functionality to the calendar view - #421
- (Possibly a stretch goal) - Enable drag and drop to reschedule a flow run - #422

src/pages/Calendar/Calendar-Day.vue
```diff
@@ -0,0 +1,393 @@
...
+          endTime: this.date,
+          flowId: this.flowId
+        },
+        loadingKey: 'loadingKey'
+      })
+      const allRuns = [
+        ...(finishedRuns?.data?.flow_run ? finishedRuns.data.flow_run : []),
+        ...(scheduledRuns?.data?.flow_run ? scheduledRuns.data.flow_run : []),
+        ...(runningRuns?.data?.flow_run ? runningRuns.data.flow_run : []),
+        ...(ongoingRuns?.data?.flow_run ? ongoingRuns.data.flow_run : [])
```
znicholasbrown, line 134:
```suggestion
        ...(finishedRuns?.data?.flow_run || []),
        ...(scheduledRuns?.data?.flow_run || []),
        ...(runningRuns?.data?.flow_run || []),
        ...(ongoingRuns?.data?.flow_run || [])
```

src/pages/Calendar/Calendar-Day.vue
```diff
@@ -0,0 +1,393 @@
...
+      fetchPolicy: 'cache-first',
+      loadingKey: 'loadingKey',
+      update: data => data.flow_by_pk
+    }
+  }
+}
+</script>
+
+<template>
+  <v-skeleton-loader
```
znicholasbrown, line 249: I wonder if we can forgo the skeleton loader in favor of showing the calendar immediately but empty? I think it would be a little bit less jarring

src/pages/Calendar/Calendar-Day.vue
```diff
@@ -0,0 +1,393 @@
...
+    },
+    handleEventClick({ nativeEvent, event }) {
+      const open = () => {
+        this.selectedEvent = event
+        this.selectedElement = nativeEvent.target
+        this.selectedOpen = true
+      }
+      if (this.selectedOpen) {
+        this.selectedOpen = false
+        setTimeout(open, 10)
```
znicholasbrown, line 208: As a matter of best practice, it'd be good to capture this timeout id and clear it at the start of this handler

src/pages/Calendar/Calendar-Day.vue
```diff
@@ -0,0 +1,393 @@
...
+  methods: {
+    async flowRunEventsList() {
+      this.gettingRuns = true
+      const scheduledRuns = await this.$apollo.query({
+        query: require('@/graphql/Calendar/calendar-day-scheduled-flow-runs.gql'),
+        variables: {
+          startTime: this.start,
+          endTime: this.end,
+          flowId: this.flowId
+        },
```
znicholasbrown, line 100: It might be cleaner to define these once at the top of the method and then reference them in each of the queries, something like this:
```vue
    const queryVariables = {
      startTime: this.start,
      endTime: this.end,
      flowId: this.flowId
    }

    const finishedRuns = await this.$apollo.query({
        query: require('@/graphql/Calendar/calendar-day-flow-runs.gql'),
        variables: queryVariables,
        loadingKey: 'loadingKey'
    })
```

We could also collect the queries using `Promises.all` to let these run in parallel instead of sequentially.

src/pages/Calendar/Calendar-Flow.vue
```diff
@@ -0,0 +1,248 @@
...
+      const ids = this.flowRuns?.map(flowRun => flowRun.flow_id)
+      return ids
+    },
+    overlay() {
+      return this.loadingKey > 0 || this.gettingRuns
+    }
+  },
+  watch: {
+    async date() {
+      this.gettingRuns = true
```
znicholasbrown, line 58: It looks like you don't need to re-allocated `this.gettingRuns` in this watcher, since `this.flowRunEventsMethod` is already doing that

src/pages/Calendar/Calendar-Flow.vue
```diff
@@ -0,0 +1,248 @@
...
+  async created() {
+    this.flowRunEvents = await this.flowRunEventsMethod()
+  },
+  methods: {
+    async flowRunEventsMethod() {
+      this.gettingRuns = true
+      const timeGroups = this.timeGroups()
+      let event = []
+      timeGroups.forEach(async (timeGroup, i) => {
+        const finished = await this.$apollo.query({
```
znicholasbrown, line 72: Same note about these methods as above - can store the variables in a single variable and re-use them and also use `Promise.all` to let them run in parallel.

src/pages/Calendar/Calendar-Flow.vue
```diff
@@ -0,0 +1,248 @@
...
+  },
+  apollo: {
+    flowRuns: {
+      query: require('@/graphql/Calendar/distinct-on-calendar-flow-runs.gql'),
+      variables() {
+        return {
+          project_id: this.projectId == '' ? null : this.projectId,
+          startTime: this.convertCalendarStartTime(this.date),
+          endTime: this.end,
+          distintOn: true
```
znicholasbrown, line 149: I think this is a typo

src/pages/Calendar/Calendar-FlowFilter.vue
```diff
@@ -0,0 +1,273 @@
...
+      },
+      loadingKey: 'loadingKey',
+      update: data => data.flow_run
+    },
+    scheduledFlowRuns: {
+      query: require('@/graphql/Calendar/calendar-scheduled-flow-runs.gql'),
+      variables() {
+        return {
+          startTime: this.start,
+          endTime: this.end
```
znicholasbrown, line 155: Can put this in a computed prop and reference them in all the queries

src/pages/Calendar/Calendar-FlowFilter.vue
```diff
@@ -0,0 +1,273 @@
...
+        return this.skip
+      },
+      loadingKey: 'loadingKey',
+      update: data => data.flow_run || []
+    },
+    allFlows: {
+      query: require('@/graphql/Calendar/calendar-flow-groups.gql'),
+      variables() {
+        return {}
+      },
```
znicholasbrown, line 196:
```suggestion
```

src/pages/Calendar/Calendar-FlowFilter.vue
```diff
@@ -0,0 +1,273 @@
...
+      return ordered
+    }
+  },
+  watch: {
+    async tenant() {
+      this.refetching = true
+      await this.$apollo.queries.flowRuns.refetch()
+      await this.$apollo.queries.scheduledFlowRuns.refetch()
+      await this.$apollo.queries.ongoingFlowRuns.refetch()
+      await this.$apollo.queries.runningFlowRuns.refetch()
```
znicholasbrown, line 103: If we need to await these, let's collect them with `Promise.all` (in the `backend` watcher as well)

znicholasbrown (changes requested):
This is looking good @zhen0! A few notes:

* [x] it's not immediately apparent why some flows in the sidebar are in blue, it sort of looks like a visual bug. Maybe we can use a small blue dot or something offset before their name to indicate that those flows have runs?
* [x] there's a small box shadow above the alert on the scheduled flow run menu
* [ ] there's a weird flickering that happens when clicking inside the flow run box, i think the event is being incorrectly propagated to the calendar element
* [x] I wrote this in the review comments as well but I'm wondering if we should remove the skeleton loader and always display the calendar instead - the reloading experience is a bit jarring. same for the flow sidebar, maybe display a small progress loader at the top rather than hiding all the flows while new ones are being fetched/calculated
* [x] the flow expansion panel is still moving when clicked, possibly a small padding inconsistency

znicholasbrown (approved):
This LGTM @zhen0 - getting this into master for some more testing! :)

Issue #407, opened 2020-11-09
Sticky Sub page nav


znicholasbrown:
**Why**

Users must scroll to the top of the page to navigate and access page-wide actions (e.g. the `Set State` or `Quick Run` actions). This can be annoying on small screens in particular.

**What**

* [x] a user can see the sub page navigation bar no matter what vertical position the page is on
* [x]  the sub page nav fluidly breaks from and joins to its place on the page and sticks to the bottom of the nav bar 

**Notes**
This may mean we have to refactor the subpage nav component to optionally include a list of tabs. The upside of this is we'll be able to remove a lot of code from the page components.

whimsicallyson:
not sure why/how this is still open

PR #441, opened 2020-11-16
Sub page nav

whimsicallyson:
PR Checklist:

From #407, moves page nav tabs to its own component as a slot within `SubPageNav`, and fixes the subpage nav directly below the navbar on pages where it exists.
- This should work on all pages where there are nav tabs (Task, TaskRun, Flow, FlowRun, Dashboard -- if there are others I haven't found them so please let me know!) 
- On pages without tabs, or on screen sizes where the tabs are at the bottom, it should properly handle the smaller size of fixed content at the top


whimsicallyson:
@znicholasbrown some code for whenever you want to take a look!

src/components/NavTabBar.vue
```diff
@@ -0,0 +1,144 @@
...
+                tb.cardLinkText
+              }}</ExternalLink
+              >!
+            </div></v-card-text
+          >
+        </v-card>
+      </v-menu>
+    </v-tab>
+    <v-spacer />
+    <v-tab
```
whimsicallyson, line 131: I'm assuming secondary tabs are largely going to be things like settings (the only one we have now) and that those won't need additional special cases like badges or explainer cards. They can always be added later if we need them...

src/components/NavTabBar.vue
```diff
@@ -0,0 +1,144 @@
...
+            </v-badge>
+          </v-card-title>
+          <v-card-text>
+            <div>
+              {{ tb.cardText }}
+              <ExternalLink :href="tb.cardLink">{{
+                tb.cardLinkText
+              }}</ExternalLink
+              >!
+            </div></v-card-text
```
whimsicallyson, line 125: This is written in a way that its pretty specific to how the Artifacts card is currently set up, but (afaict) if I try to pass in the entire message with a `v-html`, I can't use `ExternalLink`. Thoughts on which would be the better option?

src/components/NavTabBar.vue
```diff
@@ -0,0 +1,144 @@
...
+        :open-on-click="false"
+        :close-on-content-click="false"
+        offset-y
+      >
+        <template #activator="{on}">
+          <div
+            :style="tb.cardText && 'cursor: default; pointer-events: auto;'"
+            v-on="on"
+            @click.stop
+          >
```
whimsicallyson, line 94: this feels kinda hacky, but I need to override the `pointer-events:none` of a disabled tab without also making it possible to click on...

src/pages/Dashboard/Dashboard.vue
```diff
@@ -207,7 +189,17 @@ export default {
...
+      <span
+        slot="page-type"
+        style="align-items: center;
+      display: flex;
+      flex-direction: column;
+      font-size: 0.75rem;
+      height: 100%;
+      justify-content: center;
+      text-align: center;"
+        ><v-icon>view_quilt</v-icon>DASH</span
```
zhen0, line 201: Is this meant to be dashboard?

zhen0 (commented):
One question/comment from me @whimsicallyson. Otherwise, this is looking nice!

src/pages/Dashboard/Dashboard.vue
zhen0: Screen Shot 2020-11-24 at 11 11 51 AM

whimsicallyson: Nicholas and I discussed this yesterday, since all the other pages have short names (flow, task, etc) we figured Dash was ok to make the spacing more consistent :)

znicholasbrown: I think we still need to work on that spacing/design though, it's not quite there yet

src/layouts/SubPageNav.vue
```diff
@@ -21,52 +33,51 @@ export default {
...
-                  name="breadcrumbs"
-                ></slot>
-              </div>
-            </div>
-          </v-col>
-        </v-row>
+      <v-col
+        :sm="$slots['page-actions'] ? 6 : 12"
+        class="d-flex align-end"
+        :class="{}"
```
whimsicallyson, line 55: out of curiosity, what's the reason for this?

whimsicallyson:
Screen Shot 2020-11-24 at 9 15 36 PM
I think this is still out of alignment? (@znicholasbrown)

src/pages/Dashboard/Project-Selector.vue
```diff
@@ -182,10 +180,25 @@ export default {
   color: #3b8dff;
   font-weight: 500;
 }
+</style>
 
+<style lang="scss">
 .project-selector {
-  font-size: 0.9rem;
+  font-size: 16px;
```
whimsicallyson, line 187: another 'out of curiosity' -- why switch from rems to px?

src/components/NavTabBar.vue
znicholasbrown: We've resolved this with some better use of `v-if else` statements on the tabs wrapper.

src/pages/Dashboard/Dashboard.vue
znicholasbrown: The updated design was merged with #463

src/layouts/SubPageNav.vue
znicholasbrown: That was accidentally left from testing some reactive classes, removed now!

src/pages/Dashboard/Project-Selector.vue
znicholasbrown: This was swapped to match the vuetify `v-select` that we use for versions on the flow groups page - the CSS inheritance for this particular component meant we'd have to change upstream classes as well to match the context of the `v-select` OR explicitly set it here... for simplicity I chose the latter.

znicholasbrown (approved):
This is in good shape so going to merge and test in staging 👍

PR #450, opened 2020-11-18
Add support for serving the app behind a reverse proxy

m1so:
PR Checklist:

## Describe this PR

Adding support for specifying an optional path where the application would be served if it was deployed behind a reverse proxy (for example if prefect ui would sit behind an oauth proxy). 

Note that the docker container would need to be rebuilt with the correct `build-arg`. I don't think there's a way to achieve this "on the fly" as the Vue app needs to be recompiled to support the setup (links in `index.html` correctly prefixed by webpack and base url for the router) but I am opened to suggestions 👍 

### Usage
```bash
docker build . --build-arg VUE_APP_PUBLIC_PATH=/test/
```
(notice the leading and trailing slashes)


znicholasbrown:
Hi @m1so and welcome to the community! 

I have some initial thoughts on this but want to take some time to test your implementation - the code looks solid so far 😄

znicholasbrown:
Hi @m1so - sorry for the slow review here; I've been looking into this a bit and I think we _can_ make the public assets path dynamic without having to rebuild the container (I'd really like to avoid the rebuild if possible). 

Basically we can use the existing settings blob to load the public path and router base url and then modify the webpack setting by importing it separately. Linus talks about this a little in this [SO answer](https://stackoverflow.com/a/55955479/13419215). We'd need to adapt this to the way the UI is built but I think the end UX will be better than requiring a full rebuild. If you want to try to tackle that feel free, otherwise I can pick it up!

I also noticed you added a health check route to the nginx config - I haven't tested that but would accept that as a separate PR while we work on the dynamic base route.

m1so:
> Hi @m1so - sorry for the slow review here; I've been looking into this a bit and I think we _can_ make the public assets path dynamic without having to rebuild the container (I'd really like to avoid the rebuild if possible).
> 
> Basically we can use the existing settings blob to load the public path and router base url and then modify the webpack setting by importing it separately. Linus talks about this a little in this [SO answer](https://stackoverflow.com/a/55955479/13419215). We'd need to adapt this to the way the UI is built but I think the end UX will be better than requiring a full rebuild. If you want to try to tackle that feel free, otherwise I can pick it up!

Sorry for the late reply on my part. Would be great if you could pick it up, my front-end knowledge is a bit rusty and the current solution works fine for our usecase, but I can imagine that it would be a hassle to rebuild the container for other users. For more context, this feature allows up to run an oauth proxy in front of the UI and Apollo.

> I also noticed you added a health check route to the nginx config - I haven't tested that but would accept that as a separate PR while we work on the dynamic base route.

Will do. I've noticed that the Helm chart in the Server repo contains a health check for this path, although in my tests it's hitting the Vue application instead of Nginx's health check, which isn't desirable and spams the logs a lot. However, I might be wrong as I was playing around with the container built from this branch.

michcio1234:
Hi @znicholasbrown! Any chance this will be tackled in the near future? We have the exact same use case and could really use such solution.

PR #451, opened 2020-11-18
Footer

znicholasbrown:
PR Checklist:

## Describe this PR
Adds a descriptive footer to the application with helpful links to docs, community, github etc. The footer is only visible when the mouse is near the bottom of the screen.


whimsicallyson:
this is super cool! the one thing I'd add is a hover state on the logo, make it more obvious that it's also a link.

whimsicallyson:
Screen Shot 2020-11-20 at 1 31 52 PM
from slack conversation, footer also doesn't look great on small screens...

znicholasbrown:
Ok this is ready again - ended up going with an on-scroll instead of mouse move for performance and discoverability reasons (much easier to discover this way). This also exposed some existing transition issues with the dashboard tabs so I touched those up as well. Ready for another review @whimsicallyson @ThatGalNatalie

whimsicallyson:
I'm seeing a couple screens that could use some more vertical space on smaller screens:
Screen Shot 2020-12-01 at 2 46 51 PM
- flows tab of dashboard
Screen Shot 2020-12-01 at 2 49 12 PM
- teams/users/account settings tabs

are we concerned about taller screens? for example, with my typical window size, I don't have to scroll on the main dashboard so I'd never see it there...

ThatGalNatalie:
Maybe the fade in transition could be a little more faster? Primarily the dashboard but maybe it's just me

znicholasbrown:
@whimsicallyson I added a 50px buffer on the bottom scrolling to account for taller screens. Also fixed some of that weirdness when it comes to the bottom navigation by making the bottom nav components `app`-level. Should fix all of those spacings down to 350px width. Technically our tables break at around 500px so I think that should be sufficient.

@ThatGalNatalie theoretically yes, but we'd need to write a new slide-in transition instead of using the stock one that vuetify comes with.

whimsicallyson:
all looks great now! :shipit:

PR #465, opened 2020-11-25
Set State Modal 

ThatGalNatalie:
PR Checklist:

## Describe this PR
Changes the design of the Set State modal to match the design of the other forms


znicholasbrown:
This is looking really nice @ThatGalNatalie, 2 initial requests before code review:

1. The autocomplete for the optional state change reason is trying to autocomplete with a geographic state - I think we can use `autocomplete="off"` so it doesn't do that
2. Could we add a loading state to the `Confirm` button so that it can't be clicked multiple times?

src/components/SetStateDialog.vue
```diff
@@ -52,103 +52,84 @@ export default {
...
+          </div>
+
+          <v-card-actions>
+            <v-spacer></v-spacer>
+            <v-tooltip top>
+              <template #activator="{ on, attrs }">
+                <v-btn
+                  v-disable-read-only-user="!selectedState"
+                  :loading="markAsLoading"
+                  color="primary"
```
znicholasbrown, line 102: Can you make the loading spinner for this white?

src/components/SetStateDialog.vue
```diff
@@ -52,103 +52,84 @@ export default {
...
-                    Continue
-                  </v-btn>
 
-                  <v-btn text @click="reset">Cancel</v-btn>
-                </v-card-actions>
-              </v-stepper-content>
+      <v-card flat :loading="markAsLoading">
+        <div style="padding: 20px;">
+          <v-card-title class="headline word-break-normal mb-3" primary-title>
+            Change the state of {{ taskRun ? taskRun.task.name : flowRun.name }}
```
znicholasbrown, line 59: Let's differentiate the task run/flow run name here. Also I think this will read:
```
Change the state of
```
for task runs without names, so we should add a check for that and use the task name with something like "run of <<task name>>" if it's not available.

znicholasbrown (changes requested):

src/components/SetStateDialog.vue
ThatGalNatalie: okay so something like this? 

```
 Change the run of
            {{
              taskRun
                ? taskRun.task.name
                : taskRun && !taskRun.task.name
                ? taskRun.name
                : flowRun.name
            }}
```

src/pages/FlowRun/FlowRun.vue
```diff
@@ -321,6 +324,13 @@ export default {
       </v-tab-item>
     </v-tabs-items>
 
+    <Alert
+      v-if="getAlert.alertShow"
+      :value="getAlert.alertShow"
+      :type="getAlert.alertType"
+      :message="getAlert.alertMessage"
+    />
+
```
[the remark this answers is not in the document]
ThatGalNatalie: Oops my bad. Didn't realize it was registered in App. I will remove that

src/components/SetStateDialog.vue
znicholasbrown: Hm I think this:
```
 Change the run of
            {{
              taskRun ? taskRun.name ? taskRun.name : taskRun.task.name : flowRun.name
            }}
```

src/components/SetStateDialog.vue
```diff
@@ -52,103 +52,99 @@ export default {
...
+              {{ dialogType }}
+              {{
+                taskRun
+                  ? taskRun.name
+                    ? taskRun.name
+                    : taskRun.task.name
+                  : flowRun.name
+              }}
+              to
+              <span class="font-weight-black pb-8"> {{ selectedState }}.</span>
```
ThatGalNatalie, line 141: Should I also change this to the following? 
```
{{
                taskRun && taskRun.name
                  ? taskRun.name
                  : taskRun
                  ? taskRun.task.name
                  : flowRun.name
}}
```

znicholasbrown: Yup that should be fine @ThatGalNatalie !

ThatGalNatalie:
Okay I think it's good to go now unless there was something else? @znicholasbrown

src/components/SetStateDialog.vue
```diff
@@ -52,103 +52,102 @@ export default {
...
+              {{ dialogType }}
+              {{
+                taskRun && taskRun.name
+                  ? taskRun.name
+                  : taskRun
+                  ? taskRun.task.name
+                  : flowRun.name
+              }}
+              to
+              <span class="font-weight-black pb-8"> {{ selectedState }}.</span>
```
zhen0, line 144: I actually think we could take all of this out.  I know it was in the old version and the tooltip is a great addition for the  necessary warnings about child and downstream tasks but I'm not sure this section/warning adds much and possibly detracts from the more important warnings.

src/components/SetStateDialog.vue
```diff
@@ -206,4 +205,19 @@ export default {
   color: #000;
   font-weight: bold !important;
 }
+
+.checkbox-container {
+  margin-left: 30px;
+  margin-top: -20px;
+}
+/* stylelint-disable */
+.v-btn__loader {
```
zhen0, line 214: Editing this vuetify class risks affecting every button loader in the app (same for the theme light subheader above.  You can see the affect of this in the flow tile if you delete a label from this branch - the loader is not visible (but is in dev).  Elsewhere we get around this issue by placing the Vuetify class inside of a parent class - Happy to pair on this if you're not sure what I mean putting it in inside a parent class (or there's examples in other places e.g. the flow summary tile.)

zhen0 (changes requested):
This looks great - really like the new design @ThatGalNatalie  One suggestion (possibly cutting some of the tooltip wording) and one requested change (the css that affects other components) from me.

zhen0 (commented):
LGTM @ThatGalNatalie - I'll let @znicholasbrown give final approval.

znicholasbrown (approved):
LGTM @ThatGalNatalie 👍

Issue #464, opened 2020-11-25
BUG: Duration Ticker Continues to Tick for Skipped Conditional Task


AndrewHannigan:
**Bug Description**
The duration ticker continues to tick up indefinitely for a task that was skipped due to a conditional case statement.  The state of the task is correctly listed as "Skipped", but for some reason the duration ticker continues to tick.

**Steps To Reproduce**
Create a flow with a true-false conditional branch and observe the duration behavior of the skipped branch

```
import prefect                                          
from prefect import task, Flow, case, Parameter         
                                                        
@task                                                   
def hello_task():                                       
    logger = prefect.context.get("logger")              
    logger.info("Hello, Cloud!")                        
                                                        
@task                                                   
def dont_hello_task():                                  
    logger = prefect.context.get("logger")              
    logger.info("No Hello For You, Cloud!")             
                                                        
x = Parameter('say_hi', default=True)                   
                                                        
with Flow("hello-flow") as flow:                        
    x = Parameter('say_hi', default=True)               
                                                        
    with case(x, True):                                 
        hello_task()                                    
                                                        
    with case(x, False):                                
        dont_hello_task()                               
                                                        
flow.register(project_name="Test Prefect")              
```

**Browsers Tested:**
* [x] Chrome
* [ ] Firefox
* [ ] Safari
* [ ] Edge
* [ ] IE

----------

![image001](https://user-images.githubusercontent.com/3100896/100270810-3cc50000-2f1e-11eb-87ed-265f7c4a00e3.jpg)

PR #468, opened 2020-11-25
Fix schematic node finished task run durations

znicholasbrown:
PR Checklist:

## Describe this PR
Adds some conditional logic to schematic node duration calculations. This resolves cases where a task runs in finished states that have their `start_time` field populated but no `end_time` had runaway durations based on the current time.

Resolves: #464


ThatGalNatalie (approved):
looks good to me!

PR #488, opened 2020-12-04
Tokens

### LINKED ISSUE #473: BUG: New token dialog should read "Close" instead of "Cancel"
**Why**
When creating a new token, the dialog box with the token value reads "Cancel" to close the box... this indicates that pressing the button will cancel the token creation, which isn't the case. The dialog box should read "Close" after the token has been created.

**What**
* [ ] the dialog box for new tokens reads "Close" after the token has been created

zhen0:
PR Checklist:

## Describe this PR
Switches "cancel" to "close" on token dialogs. 
Closes #473


znicholasbrown (approved):
Nice 👍

Issue #477, opened 2020-12-01
Banner Directing New Users to Tutorials


dylanbhughes:
**Why**

Users who come to the Prefect Cloud UI for the first time may be unaware that there are tutorials baked in

**What**

* [ ] User signs up for Prefect Cloud
* [ ] User sees options for docs & to join slack
* [ ] Once the user arrives at the Dashboard, if they have no Flows Registered, we display a banner directing them to the tutorials available in the UI 

**Notes**

Great way to point out an excellent feature that may not get used as much as it could!

PR #516, opened 2020-12-19
tutorial banner 

ThatGalNatalie:
PR Checklist:

## Describe this PR
Adds a banner to direct new users to the tutorials page


ThatGalNatalie:
Some of the things I'm not sure about: 

1. The placement of the banner. Currently, it's sitting on top of the `SubPageNav` and gets hidden on scroll but I feel like there might be a better place for it? 
2. The component placement. It's currently a child of `SubPageNav` but I feel like it should be in the `Dashboard` comp .  I needed to know the scroll value and when to hide the banner because it will overlap with the nav if not hidden but then I end up passing props down to the `TutorialBanner` comp that are not used by the `SubPageNav` comp and it feels somewhat redundant. Could possibly use emit?

src/components/TutorialBanner.vue
```diff
@@ -0,0 +1,76 @@
...
+      query: gql`
+        query {
+          flow_aggregate {
+            aggregate {
+              count
+            }
+          }
+        }
+      `,
+      update: data => data.flow_aggregate.aggregate.count
```
ThatGalNatalie, line 44: @znicholasbrown Changed to this instead. I'm guessing this is similar to `flow_group_aggregate` but it doesn't count the archived flows?

whimsicallyson:
it works for me now! A couple thoughts, which you can take or leave as you see fit:
- the buttons don't quite line up vertically with the text, which makes sense on smaller screens (looks like two rows) but feels a bit funny on big screens because they're just barely offset
- The "dismiss" button runs riiight up to the edge of the page, I'd probably add some padding on the right side
- Can we also add a link to the tutorials in the text? In case folks don't look over to the buttons (particularly on big screens)
- I almost feel like this would be more impactful as a snackbar? That would be a big change though, so I don't blame you if you want to keep it as-is.

src/components/TutorialBanner.vue
```diff
@@ -0,0 +1,59 @@
...
+
+<template>
+  <v-banner :value="showBanner">
+    <v-avatar slot="icon" color="#3b8dff" size="40">
+      <v-icon icon="mdi-lock" color="white">
+        school
+      </v-icon>
+    </v-avatar>
+    Pst! It looks like you don't have any flows yet; check out the
+    <a href="tutorial">tutorials</a> for walkthroughs on writing and registering
```
znicholasbrown, line 40: The tutorial link should probably be using `router-link` so that we're using in-app navigation and the documentation link should open in a new tab (using the `ExternalLink` component)

znicholasbrown (changes requested):

znicholasbrown (approved):
LGTM @ThatGalNatalie 

Closes: #477

PR #524, opened 2020-12-28
 simplify and improve calendar backend and tenant switch

zhen0:
PR Checklist:

## Describe this PR
- Neatens the tenant switch logic for the calendar page so that flows from previous tenant are not shown
- Better handles the backend switch, especially when there is no server running (where previously old cached data was incorrectly shown)


znicholasbrown (approved):
LGTM @zhen0

Issue #525, opened 2020-12-29
api page gives a blank page when switching between routes 


ThatGalNatalie:
**Bug Description**
When visiting the api page for the first time, it will display the page but if you switch to a different route and head back to the api page, it gives a blank page. 

**Steps To Reproduce**
1. When arriving at the Dashboard, head to the api page
2. Click back to the dashboard or any other route
3. Head back to the api page 

**Browsers Tested:**
* [x] Chrome
* [x] Firefox
* [x] Safari
* [ ] Edge
* [ ] IE

----------
**Notes**

https://user-images.githubusercontent.com/40726391/103254530-85674300-493a-11eb-9500-8ff9be448856.mov

ThatGalNatalie:
Actually not sure if this a bug or just my settings but it displays on the first visit 

@whimsicallyson

whimsicallyson:
hm, I'm seeing this on dev.prefect as well, but there are no errors showing up in the console. not sure why this could be occurring?

PR #526, opened 2020-12-29
add beforeDestroy hook to fix blank page on reload

whimsicallyson:
PR Checklist:

## Describe this PR
Fixes issue #525

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -11,11 +11,11 @@
       http-equiv="Content-Security-Policy"
       content="
       default-src *;
-      script-src 'self' 'unsafe-eval' 'unsafe-inline' https://cdn.lr-ingest.io https://kit.fontawesome.com https://js.stripe.com https://unpkg.com;
+      script-src 'self' 'unsafe-eval' 'unsafe-inline' https://cdn.lr-ingest.io https://kit.fontawesome.com https://js.stripe.com https://unpkg.com/graphiql/graphiql.min.js https://unpkg.com/react-dom/umd/react-dom.production.min.js https://unpkg.com/react/umd/react.production.min.js;
       img-src * data:;
-      style-src-elem 'self' 'unsafe-inline' https://fonts.googleapis.com https://kit-free.fontawesome.com https://unpkg.com;
-      script-src-elem 'self' 'unsafe-inline' https://cdn.lr-ingest.io https://kit.fontawesome.com https://js.stripe.com https://unpkg.com;
-      style-src 'self' 'unsafe-inline' https://fonts.googleapis.com https://kit-free.fontawesome.com https://unpkg.com;
+      style-src-elem 'self' 'unsafe-inline' https://fonts.googleapis.com https://kit-free.fontawesome.com https://unpkg.com/graphiql/graphiql.min.css;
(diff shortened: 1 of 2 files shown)


ThatGalNatalie:
This fixes the blank page but when I try to view the api page on Firefox and Safari it won't load the page. In the console, it's giving the following errors: 
```
Content Security Policy: The page’s settings blocked the loading of a resource at https://unpkg.com/graphiql/graphiql.min.css (“style-src”).
```

```
Error: Loading CSS chunk interactive-api~851fd946 failed.
(/css/interactive-api~851fd946.73c442da.css)
```

whimsicallyson:
@ThatGalNatalie this should fix Safari at least, I'm currently seeing some breaking CSS issues on Firefox that I'm gonna dig into further. Is that what you see as well?

ThatGalNatalie:
Works for me on all the browsers. What's currently breaking?

zhen0:
Loading fine for me!  One small visual thing - the highlighting box around the run button seems to be cut off at the bottom.  (Only seems visible in Chrome?). Not a biggie and I think is also the same in dev but I now notice it every time I click that run button. 
Screen Shot 2020-12-29 at 3 46 59 PM

public/index.html
```diff
@@ -15,7 +15,7 @@
       img-src * data:;
       style-src-elem 'self' 'unsafe-inline' https://fonts.googleapis.com https://kit-free.fontawesome.com https://unpkg.com;
       script-src-elem 'self' 'unsafe-inline' https://cdn.lr-ingest.io https://kit.fontawesome.com https://js.stripe.com https://unpkg.com;
-      style-src 'self' 'unsafe-inline' https://fonts.googleapis.com https://kit-free.fontawesome.com;
+      style-src 'self' 'unsafe-inline' https://fonts.googleapis.com https://kit-free.fontawesome.com https://unpkg.com;
```
znicholasbrown, line 18: Ah, taking another look at these, I think both the style and script src content CSP lines need to be more specific for graphiql

src/pages/InteractiveAPI/InteractiveAPI.vue
```diff
@@ -93,6 +93,10 @@ query { hello }
       document.getElementById('graphiql')
     )
   },
+  beforeDestroy() {
+    const cdnScripts = ['react', 'react-dom', 'graphiql-js']
+    cdnScripts.forEach(script => document.getElementById(script).remove())
```
znicholasbrown, line 98: Is there a downside to removing these every time the component is destroyed? My understanding is that these scripts can be fairly large requests.

znicholasbrown (changes requested):
Nice @whimsicallyson - this fixes the issue but I have a couple questions on the implementation.

src/pages/InteractiveAPI/InteractiveAPI.vue
whimsicallyson: they don't seem comparatively huge,  235kb over network/875kb uncompressed. Also, this ensures we're getting the most recent versions (not likely an issue for navigating between pages, but could be if we were to check the cache for a file instead)

whimsicallyson:
>  One small visual thing - the highlighting box around the run button seems to be cut off at the bottom

made a custom focus style for that @zhen0! 
Screen Shot 2020-12-30 at 2 36 05 PM

src/pages/InteractiveAPI/InteractiveAPI.vue
znicholasbrown: Hm ok - that seems fine.

znicholasbrown (approved):
Thanks for the changes @whimsicallyson - LGTM

Issue #519, opened 2020-12-22
IAPI: Re-add URL-encoded queries


whimsicallyson:
**Why**
Our home-grown IAPI had the ability to share URL-encoded queries, and automatically populate a query from a URL. GraphiQL doesn't have this by default, but it would be nice (they may have it as an option that I didn't activate)

**What**
Find a way to implement URL-encoded queries with our new API explorer

whimsicallyson:
ref: https://github.com/graphql/graphiql/issues/1003

znicholasbrown:
Closed by: #529

PR #529, opened 2020-12-31
add support for url-encoded queries

whimsicallyson:
PR Checklist:

## Describe this PR
a page opened with a URL query should auto-populate (though not auto-run)
when a query is run, the URL will update with an encoded version of the query
#519


whimsicallyson:
someone more confident in their GQL than me should probably stress test this with tougher queries than "hello" and "user" :)

znicholasbrown (changes requested):
This is working but I'm seeing some odd behavior: when I create a query, for example:
```graphql
query {
  tenant {
    id
    name
  }
}
```

and then reload the page, I end up with this:
```graphql
{
  tenant(limit: 100) {
    id
    name
  }
}
```
which is a little odd from a user perspective. Perhaps we could URL-encode on input rather than on query?

znicholasbrown (approved):
LGTM @whimsicallyson 👍

Issue #413, opened 2020-11-10
BUG: Timezone selector has transparent background


znicholasbrown:
**Bug Description**

When navigating to the User profile page, the timezone selector dropdown sometimes has a transparent background .

**Steps To Reproduce**

Unable to consistently reproduce.

**Browsers Tested:**
* [x] Chrome
* [ ] Firefox
* [ ] Safari
* [ ] Edge
* [ ] IE

----------
**Notes**

![Screen Shot 2020-11-10 at 1 33 33 PM](https://user-images.githubusercontent.com/27291717/98716538-b921e580-2359-11eb-81e8-7e8ec09e535b.png)

This is most likely a css scoping issue with the `v-select` component or the `v-item`s that make up the dropdown selector.

PR #532, opened 2021-01-05
Flow Group schedule timezones

ThatGalNatalie:
PR Checklist:

## Describe this PR
Adds the `timezone` field to the `set_flow_group_schedule` mutation to allow the clocks to be set to the user's current timezone

### DIFF
--- a/src/pages/Flow/Settings/ClockForm.vue
+++ b/src/pages/Flow/Settings/ClockForm.vue
@@ -119,30 +119,21 @@ export default {
           </v-chip-group>
 
           <v-fade-transition mode="out-in">
-            <CronForm
-              v-if="advancedType == 'cron'"
-              key="Cron"
-              v-model="cronModel"
-              :valid.sync="valid"
-              class="mt-4"
-            />
-            <IntervalForm
-              v-else-if="advancedType == 'interval'"
-              key="Interval"
-              v-model="intervalModel"
-              class="mt-4"
-            />
+            <div v-if="advancedType == 'cron'" key="Cron">
+              <CronForm v-model="cronModel" :valid.sync="valid" class="mt-4" />
+              <v-autocomplete
+                v-model="selectedTimezone"
+                :items="tzs"
+                outlined
+                label="Time Zone"
+                style="margin-top: 110px;"
+                prepend-inner-icon="access_time"
+              />
+            </div>
+            <div v-else-if="advancedType == 'interval'" key="Interval">
+              <IntervalForm v-model="intervalModel" class="mt-4" />
+            </div>
           </v-fade-transition>
-
-          <v-autocomplete
-            v-show="advancedType == 'cron'"
-            v-model="selectedTimezone"
-            :items="tzs"
-            outlined
-            label="Time Zone"
-            style="margin-top: 150px;"
-            prepend-inner-icon="access_time"
-          />
         </div>
         <div v-else key="2" class="mt-4 d-block" style="max-width: 100%;">
           <SimpleForm v-model="simpleModel" />
--- a/src/pages/Flow/Settings/ClockForm.vue
+++ b/src/pages/Flow/Settings/ClockForm.vue
@@ -134,14 +134,16 @@ export default {
             />
           </v-fade-transition>
 
-          <!-- <v-select
-            v-if="advancedType == 'cron'"
+          <v-autocomplete
+            v-show="advancedType == 'cron'"
             v-model="selectedTimezone"
-            style="margin-top: 100px;"
             :items="tzs"
-            label="Time Zone (Optional)"
+            label="Time Zone"
+            style="margin-top: 150px;"
+            outlined
             prepend-inner-icon="access_time"
-          ></v-select> -->
+            :menu-props="{ offsetY: true }"
+          />
         </div>
         <div v-else key="2" class="mt-4 d-block" style="max-width: 100%;">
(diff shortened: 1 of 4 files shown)


src/pages/Flow/Settings/ClockForm.vue
```diff
@@ -154,3 +158,10 @@ export default {
     </div>
   </v-container>
 </template>
+
+<style>
+/* stylelint-disable */
+.v-autocomplete__content.v-menu__content .v-select-list {
+  max-width: 100%;
+}
+</style>
```
ThatGalNatalie, line 167: Not sure if we should keep this or target it through a parent class? I think it could also fix [413](https://github.com/PrefectHQ/ui/issues/413)?

znicholasbrown: Oh yeah that's a good flag, if we can target this through a parent class that would be good. You'll probably need to add the class through the `v-autocomplete` component though, with the `classList` option param

znicholasbrown: (the v-select gets attached to the body node so it's not scope-able within the component)

ThatGalNatalie: I was playing around in the inspector and tried adding this in the console `document.body.getElementByClassName('v-autocomplete__content')[0].classList.add('tz-autocomplete')` then tried styling it through the newly added class but it didn't apply. Though, I think I'm not targeting it correctly.

When you say that v-select gets attached to the body node, so the class that styles the content is outside of the component so thats why we can't just give it a parent class through `class=""`?

src/pages/Flow/Settings/Schedules.vue
```diff
@@ -141,7 +142,8 @@ export default {
               input: {
                 flow_group_id: this.flowGroup.id,
                 cron_clocks: cronClocks,
-                interval_clocks: intervalClocks
+                interval_clocks: intervalClocks,
+                timezone: this.clocks[0].timezone || this.timezone
```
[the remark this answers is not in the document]
znicholasbrown: Another note (but this has to do with the Server implementation so I'll raise there): this mutation assumes a single timezone for all clocks, which is probably not desirable. We can use this for now but I think it'll be confusing in the long run

src/pages/Flow/Settings/Schedules.vue
```diff
@@ -301,6 +303,7 @@ export default {
                     <CronClock
                       v-if="clock.type == 'CronClock'"
                       :cron="clock.cron"
+                      :timezone="clocks[0].timezone || timezone"
```
znicholasbrown, line 306: Same as above, is there a chance this fails the check?

znicholasbrown (changes requested):
I'm seeing that setting a clock's timezone doesn't persist in the display or the dropdown when I visit the schedule again. I'd expect that the timezone that's set is reflected in the schedule card (and is pre-selected when editing that schedule)

whimsicallyson:
played with this a little bit today, I think it'll make for a much nicer experience! One thing I noticed, which I don't think is great UX:

- Start without a timezone set in your user profile
- Create a cron schedule on a flow (say, 12:00pm, which because of how cron works is UTC)
- Set timezone in your user profile (say EST)
- Go back to your flow
- Schedule is now set for 12:00 EST

I'm worried that will mess up peoples' existing flow schedules? But I'm also not sure how likely of a scenario it is (I triggered it because I was surprised the scheduler wasn't defaulting to a TZ, not realizing I hadn't set one previously)

ThatGalNatalie:
> played with this a little bit today, I think it'll make for a much nicer experience! One thing I noticed, which I don't think is great UX:
> 
> * Start without a timezone set in your user profile
> * Create a cron schedule on a flow (say, 12:00pm, which because of how cron works is UTC)
> * Set timezone in your user profile (say EST)
> * Go back to your flow
> * Schedule is now set for 12:00 EST
> 
> I'm worried that will mess up peoples' existing flow schedules? But I'm also not sure how likely of a scenario it is (I triggered it because I was surprised the scheduler wasn't defaulting to a TZ, not realizing I hadn't set one previously)

I think the timezone selector defaults to local if it's not set? looks like that value  is an empty string so I added `Intl.DateTimeFormat().resolvedOptions().timeZone` to get the local timezone if a timezone was not set which I think might solve that?

src/pages/Flow/Settings/ClockForm.vue
```diff
@@ -49,6 +57,9 @@ export default {
       return this.advanced ? `${this.advancedType}Model` : 'simpleModel'
     }
   },
+  beforeMount() {
+    this.selectedTimezone = this.timezone
+  },
```
znicholasbrown, line 62: Couldn't we set this in the data ref instead?

znicholasbrown (changes requested):
Hm I'm not sure exactly what's changed but this doesn't seem to be working for me, at least from a display standpoint. After creating a schedule with a timezone (for example set to CET with Africa/Algiers selected), the timezone doesn't appear in the schedule display card. After refreshing, the timezone shows up but it's incorrect, set to my local timezone instead. 



A small display note: this list still contains underscores instead of spaces; we should remove those or use a different timezone list (moment may already have a list). This is true on the user settings page as well so we can fix this in a follow up PR.
Screen Shot 2021-01-22 at 11 47 12 AM

znicholasbrown (approved):
LGTM @ThatGalNatalie !

PR #556, opened 2021-01-19
Dev => Master

znicholasbrown:
PR Checklist:

## Describe this PR
This brings the new Okta-based auth changes from dev to master, and includes a new Okta-specific auth store and the relevant unit tests, as well as several new pages related to an in-app login experience. For now the in-app login experience is disabled in favor of our universal portal.


src/main.js
```diff
@@ -1,12 +1,17 @@
...
+
 import Router from 'vue-router'
 import App from '@/App.vue'
 
 // Plugins
 import store from '@/store'
 import { defaultApolloProvider } from '@/vue-apollo'
+
+// eslint-disable-next-line
+// import okta from '@/plugins/okta'
```
whimsicallyson, line 13: do we not need this to be active?

znicholasbrown: Nope, this is specifically for the in-app login experience

whimsicallyson (approved):
lgtm!

whimsicallyson:
one tiny comment - it does feel a little weird that when you log out, the header disappears before the rest of the UI

src/pages/Auth/Auth.vue
```diff
@@ -0,0 +1,114 @@
...
+      this.previousHeight = getComputedStyle(element).height
+    },
+    enter(element) {
+      const { height } = getComputedStyle(element)
+      element.style.height = this.previousHeight
+
+      setTimeout(() => {
+        element.style.height = height
+      })
+    }
```
ThatGalNatalie, line 22: Just a question, what would be the difference if you didn't put it in the `setTimeout`? As this doesn't have a delay, I would imagine it would run immediately?

znicholasbrown: Yeah there are a few reasons we might want to do this (and to be honest I stole this from animations I wrote for the universal login application about a year ago so I can't remember which reason this was for, likely all of them):
1) Using a null timeout allows us to put this piece of logic in the event queue but _not_ execute it immediately
2) It enables next-tick-like css animations here, being that line 17 is caught by the renderer, executed, and then line 20 is rendered after. Without the timeout, line 17 is ignored by the renderer.

zhen0 (approved):
Thanks for the changes @znicholasbrown. LGTM!

PR #563, opened 2021-01-21
Add Flow Descriptions

### LINKED ISSUE #535: Feature: flow documentation
**Why**

Documentation is always nice 👍  this feature proposes to allow to document a flow and see it visible in the UI. For instance under the info block. Preferably, it would be nice if it was possible to write full markdown but a plain format string to document the flow would be good start. I believe Airflow also has this feature if the team needs some inspiration 😬 

image

**What**

* [ ] Design UI where to show documentation
* [ ] Update the UI
* [ ] Update database to store documentation string
* [ ] Update Flow constructor to pass in documentation

**Notes**

5 peeps are already hyped: https://prefect-community.slack.com/archives/CL09KU1K7/p1610014772362000

zhen0:
PR Checklist:

## Describe this PR
- Adds an editable description tab to the flow page.  Users can create a flow group description/documentation using mark down to be stored in the flow group description field of the database.  The tab will also show individual flow descriptions (currently only added at the flow.serialize step.)
- Removes the versions tab from the flow page - versions are still displayed in the Action bar of the flow page so they extra tab seemed redundant
- Closes #535 
- Draft docs update in https://github.com/PrefectHQ/prefect/pull/4002


whimsicallyson:
I'm a bit confused on what can have a description. Can a version of a flow have a different description than the group? If not, why can I edit/create a description from a specific version, it appears to just override the group description, which is not what I would have expected. And if the group is the only thing that can have a description, why don't I see that regardless of version?

Also, I would personally add a white background to the text editor (give it some differentiation) and auto-focus on clicking the edit button, if possible.

znicholasbrown (approved):
LGTM @zhen0 ! :)

PR #574, opened 2021-01-27
Refresh token improvements

znicholasbrown:
PR Checklist:

## Describe this PR
Fixes issues with tokens not being refreshed properly and improves handling of refresh mutations so in-flight queries will be respected (the `refresh_token` route won't be called again if another is already in flight).

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -8,7 +8,7 @@
 
 ### Bugfixes
 
-- None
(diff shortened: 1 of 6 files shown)


zhen0 (approved):
LGTM @znicholasbrown - We can trial in dev/staging to make sure this fixes the refresh before we release.

PR #577, opened 2021-01-28
Use role_detail instead of role

jlowin:
The `role` field in graphql is deprecated. This PR replaces calls to that field with a `role_detail` relationship to get additional information (in this case, the role name).

(Note: once released, we will remove the `role` field entirely as the UI is its only consumer and rename the `role_detail` relationship back to `role`)

PR Checklist:


znicholasbrown (approved):
LGTM @jlowin - approving but holding off on the merge pending some other changes.

PR #579, opened 2021-01-28
Dev

zhen0:
PR Checklist:


ThatGalNatalie:
Styling wise, a couple of things I noticed:

Onboarding:
- /plan the cards are not very responsive 

Account:
I think this was a part of the account page?
- is that search icon/magnifying glass icon supposed to be there? 
- the spacing  in the error alert in the plan comparison to me feels off
- the selected plan doesn't feel elevated enough to me
- I feel like there is a tiny gap between the cards from the border radius?
- in the plan comparison,  should the confirm button be  below the Add payment information
 card?  
- if I'm on a plan that I already have and I click on change plan but  choose the same one that I have, maybe it can display a different error message?

PR #585, opened 2021-02-02
Run configs and run tab update

znicholasbrown:
PR Checklist:

## Describe this PR
Opening this up for comments (netlify preview should work!)

This PR introduces run configs to the UI and is a general VU on the run tab form. It includes some updates to the following features:
- JSON inputs - adds Prefect-flavored syntax highlighting and makes the component easier to drop into other components
- Run names - names are now auto-generated in the UI (and can be randomized at-will before submitting!)

... and introduces some new components:
- new date time selector (`DateTimeSelector`) - the existing one is visually and functionally more complex than necessary and was proving too difficult to adapt to a different environment for a variety of reasons. The new one shouldn't make any assumptions about its parent container (it's no longer assumed to be a menu) and has much-simplified logic based on a single moment object
- YAML input component (`YamlInput`) - includes syntax highlighting as well!
- Dict input component - includes either simplified key/value inputs or an advanced JSON editor mode
- List input component (`ListInput`) - will be a simplified version of the current labels component, meant for more general `List[string]` use
- MultiLine input component (`MultiLineInput`) - a wrapper component that allows swapping between JSON and YAML inputs; could be easily adapted for other languages or types

TODO
- [x] Create `ListInput` component
- [x] add a run button 😆 
- [x] Float labels to the top of the run config form: labels are currently the last thing on the form but are probably one of the most-used
- [x] default no run config to... null? universal? local? I'm thinking universal so that labels and environment variables can still be passed without duplication
- [x] shorten descriptions? I'm not sure on this - some of the descriptions are a little lengthier than might be desired but maybe not
- [ ] add JSON/advanced preferences to user settings (ease-of-use, not required)
- [x] move run config form to an "advanced" part of the run form? maybe a "Show advanced" at the bottom?
- [x] add timezone selector to new datetime component
- [ ] test test test


whimsicallyson:
this is so much more intuitive, I love it! two nitpicks that I noticed:
- I set the schedule to a time in the future, but didn't run the flow. After that time passed, it was still shown in the UI as the scheduled time. What's going to happen if I click the button in that case? Will it default to immediately run? In an ideal world I'd like to see it go back to tracking the system time as if I'd never touched it, but that might be prohibitively hard.
- You used more duotone icons :(

src/components/CustomInputs/YamlInput.vue
```diff
@@ -0,0 +1,247 @@
...
+  },
+  data() {
+    return {
+      // Line that contains a syntax error
+      errorLine: null,
+      // The value may need to be modified for formatting.
+      // Store the value prop's value internally as data so it can be modified.
+      internalValue: this.value,
+      error: null,
+      focussed: false
```
whimsicallyson, line 66: was jenny here? 😉

src/components/RunConfig/DateTimeSelector.vue
```diff
@@ -0,0 +1,387 @@
...
+        case 'met':
+          icon = 'fad fa-globe-europe'
+          break
+
+        case 'universal':
+          icon = 'fad fa-planet-ringed'
+          break
+
+        case 'antarctica':
+          icon = 'fad fa-snowman'
```
whimsicallyson, line 122: 😂

src/utils/runConfigs.js
```diff
@@ -0,0 +1,287 @@
...
+    }
+  ]
+}
+
+export const KubernetesRun = {
+  type: 'KubernetesRun',
+  label: 'Kubernetes',
+  description: 'Run the flow as a job on a Kubernetes agent',
+  ref: 'https://docs.prefect.io/api/latest/run_configs.html#kubernetesrun',
+  icon: 'fak fa-kubernetes',
```
whimsicallyson, line 82: we probably want to change this to our local font, since we know we can't use the kit info for uploading icons anymore 😞

znicholasbrown: Yeah definitely, I'll tackle that conversion later today >:(

zhen0 (approved):
lgtm @znicholasbrown !

PR #593, opened 2021-02-04
Show label warning on Server

zhen0:
PR Checklist:

## Describe this PR
- When we added the label warning, agent labels weren't yet in server.  Now they are so we should enable this component in server too.  Should help reduce the questions about labels in the community slack too.


znicholasbrown (approved):
Nice 👍

PR #609, opened 2021-02-13
Dont specify labels on run config - results in bad config schemas

cicdw:
This PR fixes an issue with the Run page - anytime you use that page and don't edit the run config, we were creating a bad run config schema that couldn't be parsed by agents, resulting in flow run failures.


cicdw:
Closing in favor of #610

PR #631, opened 2021-02-24
Redirect links

ThatGalNatalie:
PR Checklist:

## Describe this PR
Aims to allow you to open up multiple links at the same time without redirecting to the first link

Before it was setting the route of the first clicked link to `localstorage` then using that as the route for the remaining  links that were clicked on


znicholasbrown:
I wonder if it might make more sense to use session storage instead of local storage for this? The key will still exist when a tab is duplicated (since duplicating tabs also duplicates anything in session storage) but I think that would be intended behavior. @ThatGalNatalie

ThatGalNatalie:
yeah I think session storage makes more since it will not overwrite it

PR #633, opened 2021-02-27
State table

ThatGalNatalie:
PR Checklist:

## Describe this PR
Enables you to sort by state in the task run table

### DIFF
--- a/src/graphql/FlowRun/table-task-runs.gql
+++ b/src/graphql/FlowRun/table-task-runs.gql
@@ -1,7 +1,7 @@
 query TableTaskRuns(
   $flowRunId: uuid
   $name: String
-  $state: String
+  $state: [String!]
   $limit: Int
   $offset: Int
   $orderBy: [task_run_order_by!]
@@ -10,7 +10,10 @@ query TableTaskRuns(
     id
     name
     task_runs(
-      where: { task: { name: { _ilike: $name } }, state: { _eq: $state } }
+      where: {
+        task: { name: { _ilike: $name } }
+        details: { state: { _in: $state } }
+      }
       order_by: $orderBy
       limit: $limit
       offset: $offset
--- a/src/graphql/FlowRun/table-task-runs.gql
+++ b/src/graphql/FlowRun/table-task-runs.gql
@@ -10,17 +10,19 @@ query TableTaskRuns(
     id
     name
     task_runs(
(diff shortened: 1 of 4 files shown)


src/graphql/FlowRun/table-task-runs.gql
```diff
@@ -9,7 +10,7 @@ query TableTaskRuns(
     id
     name
     task_runs(
-      where: { task: { name: { _ilike: $name } } }
+      where: { task: { name: { _ilike: $name } }, state: { _ilike: $state } }
```
znicholasbrown, line 13:
```suggestion
      where: { task: { name: { _ilike: $name } }, state: { _ilike: $state } }
```

I don't think we need to do a fuzzy search here, which I think will be slower; we can use `_eq` for state matching instead.

src/pages/FlowRun/TaskRunTable-Tile.vue
```diff
@@ -136,6 +139,35 @@ export default {
...
+        :items="[
+          { header: 'Finished States' },
+          { divider: '...' },
+          'Failed',
+          'Success',
+          'Cancelled',
+          'Finished',
+          'Skipped',
+          'TimedOut',
+          { header: 'Scheduled States - To Re-run Task Run' },
```
znicholasbrown, line 161:
```suggestion
          { header: 'Scheduled States - To Re-run Task Run' },
```
I'm not sure this header makes sense for a sort label?

src/pages/FlowRun/TaskRunTable-Tile.vue
```diff
@@ -136,6 +139,35 @@ export default {
...
+          { divider: '...' },
+          'Scheduled',
+          'Resume',
+          { header: 'Pending - to clear the state' },
```
znicholasbrown, line 165: I'm not sure this header makes sense for a sort label?

znicholasbrown (changes requested):
Recognizing this is in draft but I let a couple quick comments!

src/pages/FlowRun/TaskRunTable-Tile.vue
ThatGalNatalie: oh my bad, I forgot to change that. I copied it from the set state modal. For the sort, I don't think we would need headers?

znicholasbrown (approved):
This looks great @ThatGalNatalie !

PR #645, opened 2021-03-04
Flow run state filter

ThatGalNatalie:
PR Checklist:

## Describe this PR
Enables you to filter by state in the flow run table


src/pages/Flow/FlowRunTable-Tile.vue
```diff
@@ -46,7 +46,30 @@ export default {
       page: 1,
       searchTerm: null,
       sortBy: 'scheduled_start_time',
-      sortDesc: true
+      sortDesc: true,
+      state: [],
+      states: [
```
znicholasbrown, line 51: We should sort these here and on the task run tile; since there's no text input it's a little difficult atm to find the one you're looking for

znicholasbrown (changes requested):
A few comments related to sorting and some invalid states @ThatGalNatalie

PR #650, opened 2021-03-09
Mapped case task runs

ThatGalNatalie:
PR Checklist:

## Describe this PR
When a case task runs, it doesn't have a `end_time` populated in the database so the flow run timeline will show an indefinite task run duration and a skewed timeline. This aims to provide a `end_time` if it's `null`


znicholasbrown (approved):
LGTM @ThatGalNatalie 👍

PR #656, opened 2021-03-10
Dark mode!!

whimsicallyson:
PR Checklist:

## Describe this PR
Creates a dark mode/theme for the UI! I've done my best to try and find every page/popover/etc to make sure everything's legible, but I don't guarantee I've succeeded, so I would appreciate manual QA from folks.

Toggle the light/dark button in the user menu!

Open questions:
- the alt prefect and primary colors were defined before my time, are they set in stone? I think primary should be a bit lighter for links in small text size
- do we want to standardize the colors used for code syntax highlighting? the API is different from the JSON input, which I believe is different from the code samples in the tutorials


znicholasbrown (approved):
Great stuff @whimsicallyson ! 😄

PR #662, opened 2021-03-12
Named routes 

ThatGalNatalie:
PR Checklist:

## Describe this PR
This aims to change all the non dynamic name routes to a string

### DIFF
--- a/src/components/Agents/Agents.vue
+++ b/src/components/Agents/Agents.vue
@@ -470,7 +470,7 @@ export default {
         to understand their role in your infrastructure.
         <span v-if="isCloud">
           You can also go through the
-          <router-link class="link" :to="{ path: 'tutorial/Universal-Deploy' }">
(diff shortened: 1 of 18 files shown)


ThatGalNatalie:
Tried not binding the `to` and it wasn't working for me and wasn't sure why? Isn't `:to="'route'"` the same thing?

znicholasbrown:
> Tried not binding the `to` and it wasn't working for me and wasn't sure why? Isn't `:to="'route'"` the same thing?

I think not - adding `:` tells the Vue engine to do something with that attribute and in this case it's a special attribute that is calculated and transformed onto a new link element in the form of `href`; since Vue doesn't parse attributes without `v-` or `:` prefixes, this means that the attribute is interpreted literally, as `to`, which doesn't have any meaning.

znicholasbrown (approved):
LGTM @ThatGalNatalie

Issue #686, opened 2021-03-22
Replace delete flows mutation with delete_flow_group


zhen0:
**Why**

We currently loop over flows to delete more than one version from a flow group. This can bump into problems with large flows. There is now a delete_flow_group mutation which should handle big deletions more elegantly. 
**What**

* [ ] The delete flow mutation is replaced with delete_flow_group and users can delete flows with multiple versions

**Notes**

PR #689, opened 2021-03-23
delete entire flow group with new mutation

whimsicallyson:
PR Checklist:

## Describe this PR
Remove (confusing) ability to delete individual flow versions, replace deleting flow group with new `delete_flow_group` mutation. Solves #686

### DIFF
--- a/src/components/DeleteFlowButton.vue
+++ b/src/components/DeleteFlowButton.vue
@@ -23,7 +23,6 @@ export default {
     return {
       deleting: false,
       errorMessage: '',
-      allFlows: false,
       deleteDialog: false,
       // Alert data
       alertShow: false,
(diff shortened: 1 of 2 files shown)


cicdw (commented):
I don't see a benefit at this time to removing the ability to delete a single version -- it's possible some users have relied on this for deleting really old versions.  I would prefer if we kept the original logic, but for the "delete all versions" we call this new route instead, at least until we _actually_ remove versions in the backend

src/components/DeleteFlowButton.vue
```diff
@@ -34,33 +33,14 @@ export default {
...
-              success
-            }`
-        })
-        const newString = gql`
-          mutation {
-            ${string}
+      return gql`
+        mutation($flowGroupId: UUID!) {
+          delete_flow_group(input: { flow_group_id: $flowGroupId }) {
+            success
```
znicholasbrown, line 40: Since we don't need to dynamically build this mutation anymore, we can use a `require` call with a static mutation file

znicholasbrown (changes requested):
LGTM @whimsicallyson  - just 1 comment!

src/components/DeleteFlowButton.vue
whimsicallyson: since we decided we do want the ability to delete versions, I kept it as is (and added the other mutation in a conditional)

znicholasbrown (approved):
Thanks for the changes @whimsicallyson

PR #690, opened 2021-03-23
Improve authorization expiration handling

znicholasbrown:
PR Checklist:

## Describe this PR
Adds a new terminal link to the apollo client chain that checks for a pre-set authorization expiration on all requests; requests whose authorization headers have expired aren't forwarded to the batch link.

Note: Given that almost-expired tokens are still valid I decided to go with a strict check instead; a future PR could add a buffer to the check (`almostExpired`?) to account for network latency but I'm not sure if this violates any assumptions on token handling.


whimsicallyson (approved):
LGTM!

PR #692, opened 2021-03-25
Usage displays/updates

znicholasbrown:
PR Checklist:

## Describe this PR
- Adds a tenant id filter to the usage timeline and usage today tiles on the Account page
- Updates some typography on the dash (not every word needs to be capitalized 😛 )
- Adds 3 new tiles to the dashboard:
  - Upgrade (for accounts on legacy plans)
  - Free / Monthly usage (for Free, Starter, Standard, and non-committed Enterprise accounts)
  - Committed usage (for pre-purchased run packages)
-  Adds next invoice/current balance (for Starter, Standard, and non-committed Enterprise accounts) tile to the accounts page
- Adds a plan cancellation button to the account page, which allows downgrading an account to the free tier (Self-serve users only)

I also reworked the layout of the dashboard to use CSS grid; this'll make placement of tiles much easier than using a layout slot-based system (this was causing a lot of problems trying to add a half-height tile)


dylanbhughes:
CSS grid++

src/pages/Dashboard/UsageTiles/CycleUsage-Tile.vue
```diff
@@ -0,0 +1,172 @@
...
+      if (!this.invoice) return null
+      const nextPayment = this.invoice.next_payment_attempt * 1000
+      return this.formatLongDate(nextPayment)
+    },
+    periodStart() {
+      if (!this.invoice) return null
+      return new Date(this.invoice.period_start * 1000)
+    },
+    freeUsage() {
+      if (Number.isNaN(this.usage)) return null
```
ThatGalNatalie, line 29: Sometimes when switching between the plans, this.usage would become undefined and I would get "NaN% of free runs used" so I was wondering why we wouldn't use isNaN? 



NaN

znicholasbrown: Hm good point, I think `isNaN` is more appropriate here.

ThatGalNatalie (approved):

PR #706, opened 2021-03-29
Add remaining states to the state filter 

ThatGalNatalie:
PR Checklist:

## Describe this PR
This should add the remaining states to the state filter


src/utils/states.js
```diff
@@ -50,21 +50,26 @@ export const STATE_NAMES = [
   'All',
   'Success',
   'Failed',
+  'Finished',
   'Scheduled',
+  'ScheduledAlt',
```
znicholasbrown, line 55: Where does this state come from?

src/utils/states.js
```diff
@@ -50,21 +50,26 @@ export const STATE_NAMES = [
...
   'Pending',
   'Running',
   'Cancelled',
   'Cancelling',
+  'Cached',
   'Retrying',
   'Queued',
   'Paused',
+  'InProgress',
```
znicholasbrown, line 64: This is a derived state, so shouldn't be included here

src/utils/states.js
```diff
@@ -50,21 +50,26 @@ export const STATE_NAMES = [
...
   'Mapped',
   'Looped',
+  'Listening',
```
znicholasbrown, line 67: I don't think we expose this state anywhere at the moment, so let's remove this for now

znicholasbrown (changes requested):

src/utils/states.js
ThatGalNatalie: In the `states.js` file then in the `STATE_COLORS`. Should this be removed? I was also referring to this [diagram](https://docs.prefect.io/api/latest/engine/state.html#state-2)

whimsicallyson: this is there so we can reference the color for scheduled flows (its an alternating yellow stripe). If this is a bad place for it, we can move it to a var in vuetify.js instead

znicholasbrown: Hm yeah - this isn't the correct place for it, since it'll be included in state lists (and `ScheduledAlt` isn't a valid state)

src/utils/states.js
```diff
@@ -2,7 +2,6 @@ export const STATE_COLORS = {
   Failed: '#eb0000',
   Pending: '#b9dcff',
   Scheduled: '#ffbe1e',
-  ScheduledAlt: '#ffeec4',
```
znicholasbrown, line 5: It's ok if we have this here since it's for styling and isn't displayed in any list - can probably revert this line and the changes to `vuetify.js` (unless there's a different complication here i'm not aware of)

znicholasbrown (approved):
LGTM @ThatGalNatalie

PR #711, opened 2021-03-30
Disable the automations tab

znicholasbrown:
PR Checklist:

## Describe this PR
Disables the automations tab while the feature is being tested.

### DIFF
--- a/src/styles/global.scss
+++ b/src/styles/global.scss
@@ -298,6 +298,12 @@ $percents: 0, 10, 20, 30, 40, 50, 60, 70, 80, 90, 100;
   }
 }
 
+/* stylelint-disable */
+.disable.v-badge--inline .v-badge__badge,
+.v-badge--inline .v-badge__wrapper {
+  pointer-events: none;
+}
+
 .v-text-field-input-color {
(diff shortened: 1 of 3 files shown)


src/styles/global.scss
```diff
@@ -298,8 +298,13 @@ $percents: 0, 10, 20, 30, 40, 50, 60, 70, 80, 90, 100;
   }
 }
 
+/* stylelint-disable */
+.disable.v-badge--inline .v-badge__badge,
+.v-badge--inline .v-badge__wrapper {
+  pointer-events: none;
```
whimsicallyson, line 304: I'm a little surprised we need this rule, I'd expect the disabled prop to take care of it, but I trust that its here because its actually needed :)

whimsicallyson (approved):
coming soon! yay!

src/styles/global.scss
znicholasbrown: Yeah it looks like the disabled prop doesn't actually disable the badge (it's a hidden API on that component) but adds the disabled class still, so pointer events are still being propagated.

PR #714, opened 2021-03-31
check if flow version is archived for restart button state

### LINKED ISSUE #446: BUG: Restart Flow Run is not disabled for archived versions
**Bug Description**

Because we can't run an archived version of a flow, we should not be able to restart an archived flow version's runs.  The restart button should check a flow run's flow version and only restart runs from a live flow. 

**Steps To Reproduce**

1. Find a flow with several versions.  
2. Choose an old version.
3.  Restart should be disabled but is not.  

**Browsers Tested:**
* [ ] Chrome
* [ ] Firefox
* [ ] Safari
* [ ] Edge
* [ ] IE

----------
**Notes**

whimsicallyson:
PR Checklist:

## Describe this PR
Fixes #446, only allows you to click the restart button on a flow run if it isn't an archived version.


znicholasbrown (changes requested):
I don't think we need to add a new query for the flow in this component, since the flow object is being passed with the `flowRun` prop; adding the `archived` field there is better in this case.

whimsicallyson:
ugh I can't believe I missed that!

src/graphql/Flow/flow-by-pk.gql
```diff
@@ -1,9 +1,10 @@
...
+query FlowByPk($flowId: uuid!) {
+  flow_by_pk(id: $flowId) {
+    id
+    name
+    version
+    schedule
+    flow_group_id
+    archived
+  }
+}
```
znicholasbrown, line 10: Can we remove this change as well?

znicholasbrown (approved):
LGTM 👍

PR #716, opened 2021-03-31
bugfix to show helpful error message on failed service account creation

whimsicallyson:
PR Checklist:

## Describe this PR
Screen Shot 2021-03-31 at 3 06 16 PM
fix this error message

### DIFF
--- a/src/pages/TeamSettings/Service-Accounts.vue
+++ b/src/pages/TeamSettings/Service-Accounts.vue
@@ -78,7 +78,7 @@ export default {
         })
         .catch(({ graphQLErrors }) => {
           this.accountCreationError = `There was an error creating your service account:
-            ${{ graphQLErrors }}`
+            ${graphQLErrors[0].message}`
           return { error: graphQLErrors }
         })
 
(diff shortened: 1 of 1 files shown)


src/pages/TeamSettings/Service-Accounts.vue
```diff
@@ -78,7 +78,7 @@ export default {
         })
         .catch(({ graphQLErrors }) => {
           this.accountCreationError = `There was an error creating your service account:
-            ${{ graphQLErrors }}`
+            ${graphQLErrors[0].message}`
```
znicholasbrown, line 81: Do we know there will always be a message at the first index here or should this be?
```suggestion
            ${graphQLErrors?.[0].message}`
```

znicholasbrown (changes requested):

src/pages/TeamSettings/Service-Accounts.vue
whimsicallyson: I think there should likely be, but better to be safe.

znicholasbrown (approved):
Thanks for the changes!

PR #719, opened 2021-04-01
Add the proper permissions to role-based invitations

znicholasbrown:
PR Checklist:

## Describe this PR
Disables role-based controls for non-enterprise licenses (this was previously erroneously enabled, leading to errors when setting non-admin roles); this includes when inviting or modifying users.


ThatGalNatalie:
After closing the dialog and then opening it up again, the label displays "Role" instead of "Administrator" and the send button is disabled. 

role-label

znicholasbrown:
Good catch - we were completely resetting that form to `null` values when the dialog box was closed, instead of resetting to the initial state. Fixed now 👍

ThatGalNatalie (dismissed):
LGTM

PR #721, opened 2021-04-01
Small fixes to onboarding

znicholasbrown:
PR Checklist:

## Describe this PR
Fixes 2 issues:
- sometimes the full animation for the team name screen wasn't playing, leading to an incorrect y-offset on the form
- we weren't doing a full redirect of the page after license creation, which meant the cache was holding a null-license

### DIFF
--- a/src/pages/Onboard/Resources.vue
+++ b/src/pages/Onboard/Resources.vue
@@ -225,7 +225,7 @@ export default {
       >
         <v-col cols="12" class="text-center py-2">
           <v-btn
-            :to="{ name: 'dashboard' }"
+            href="/"
             class="mx-auto"
             color="primary"
             x-large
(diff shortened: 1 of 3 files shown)


ThatGalNatalie (approved):
LGTM

src/pages/Onboard/Resources.vue
```diff
@@ -225,7 +225,7 @@ export default {
       >
         <v-col cols="12" class="text-center py-2">
           <v-btn
-            :to="{ name: 'dashboard' }"
+            href="/"
```
ThatGalNatalie, line 228: Just had a question, what is the difference between using :to="{path: '/'}" and href="/" ?

znicholasbrown: Using the former handles the navigation through the router; the later is a full page refresh to the root directory. Doing this forces a refresh of the application (it's equivalent to retyping cloud.prefect.io into the url bar) and ensures local state is cleared after the license has been created.

PR #722, opened 2021-04-02
Usage timeline bugfixes

znicholasbrown:
PR Checklist:

## Describe this PR
Fixes:
- Issues with overlapping bars; bar positions were being calculated based on fixed width buckets on a continuous time scale 🙅 
- Issues where the invisible interaction areas were overlapping
- The bars no longer attempt to set negative heights caused by the chart updating before dimensions had been set; we're no longer using `requestAnimationFrame` for the initial resize call in the `mounted` hook, and we're calling the raw method instead of the throttled one to prevent the race condition altogether
- Fixes an issue where period boundaries weren't quite what we wanted them to be; months should now span only the days in the month, and weeks should go from Mon-Sun


ThatGalNatalie (approved):
lgtm

PR #736, opened 2021-04-05
Display flow version in flow run 

ThatGalNatalie:
PR Checklist:

## Describe this PR
This will add a label specifying which flow version the flow run has spawned from


whimsicallyson (dismissed):
this will mesh nicely with my change in #714

src/pages/FlowRun/Details-Tile.vue
```diff
@@ -218,6 +218,20 @@ export default {
           <v-list-item dense>
             <v-list-item-content>
               <v-list-item-subtitle class="text-caption">
+                <v-row no-gutters>
+                  <v-col cols="6">
+                    Version
```
znicholasbrown, line 223: Since runs have a version field as well, I think it'd be clearer to say something like this instead:
```suggestion
                    {{ flowRun.flow.name }} version
```

znicholasbrown (changes requested):

znicholasbrown (approved):
LGTM 👍  thanks for the changes!

PR #737, opened 2021-04-05
Make "current" label more legible

ThatGalNatalie:
PR Checklist:

## Describe this PR
The "current" label is currently illegible in both dark and light mode so this aims to make it more legible by changing the class


whimsicallyson (dismissed):
yup this'll do it :)

PR #752, opened 2021-04-08
Allow flows to have a null project

madkinsz:
PR Checklist:


whimsicallyson:
in an ideal world, we'd like this to also be able to create a dashboard page for the "no project project" and add that functionality into `Project-Selector.vue`. From a practical standpoint, though, I haven't figured out a way to do so -- we use the project ID as the route param to pull the data for the page, and I'm not sure its possible to do that with a project ID of null.

PR #755, opened 2021-04-09
licensing and permissions

ThatGalNatalie:
PR Checklist:

## Describe this PR

### DIFF
--- a/src/components/ConcurrencyInfo.vue
+++ b/src/components/ConcurrencyInfo.vue
@@ -8,10 +8,7 @@ export default {
     }
   },
   computed: {
-    ...mapGetters('license', ['license']),
-    flowConcurrency() {
-      return this.license?.terms?.flow_concurrency
-    }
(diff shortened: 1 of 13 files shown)


znicholasbrown (changes requested):
As part of the broader license store, many of these properties are unclear, e.g. `isEligible`. If we're adding these to the store I think the properties should be as homogenous as possible in access verbiage and functionality. There's a potential to include this in 1 or 2 getters that look something like this:
```
permission: state => (operation, ref) => {
    // where operation is one of (create, read, update, delete, feature)
    // and ref is the feature reference
    return state.permissions?.includes(`${operation}:${ref}`)
  }
}
```

In addition, it looks like some properties were copied directly from the components, which means some have loading states and some don't. Do we need loading states at all for these? If so, all permissions should likely return them.

ThatGalNatalie:
closing this pr in favor of a new one

Issue #649, opened 2021-03-08
Simplify navigation from flow run page to latest version of flow


jcrist:
Developing a flow against prefect cloud may require some iteration, with the following common workflow:

1. Register flow
2. Click "quick run" to kick off a flow run, navigating you to a flow run page
3. Watch the flow run progress until an error occurs
4. Edit the local code to fix the bug
5. Re-register the flow (bumping the flow version)
6. Kick off a new flow run with the new version

Unfortunately navigating from the failed flow run page to the flow page for the latest flow requires two UI actions:

- Click the flow name in the breadcrumbs in the upper left corner. This brings you to the flow page for the now outdated version.
- Navigate to the latest version of the flow in the "Version" dropdown in the upper right corner, and navigate to the latest version.

After a few iterations, this process becomes annoyingly slow. It would be nice to provide a single UI interaction to get from a flow run page to the latest version of the flow.

I'm not sure if changing the flow name link to navigate to the latest version makes the most sense, or if another UI element should be added. I personally never feel the need to look at old *flow* pages for previous versions of a flow, so navigating to that page seems less useful to me, but others may feel differently.

znicholasbrown:
Closed by #759

PR #759, opened 2021-04-09
add breadcrumb to flow version, change flow name breadcrumb to overview

whimsicallyson:
PR Checklist:

## Describe this PR
Addresses #649 by adding an additional breadcrumb. Flow name will always go to overview/latest version, version crumb will go to that specific version. Also helps make it more obvious what version of a flow a run is attached to, in case folks needed that (though 736 does very similar...).

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -4,6 +4,7 @@
 
 ### Features and Improvements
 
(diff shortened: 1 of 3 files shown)


znicholasbrown (approved):
LGTM 👍 !

Issue #718, opened 2021-04-01
Create a mixin to handle licensing and permissions


znicholasbrown:
**Why**

We currently check licenses and permissions in most of our components/pages but in different ways. Centralizing / standardizing this will reduce the maintenance burden of license/permission checks and prevent us from needing to add this functionality for every new feature.

**What**

* [ ] a license/permission mixin exists which allows us to easily check for the ability to perform certain actions/view certain screens

**Notes**

znicholasbrown:
This may not require a mixin; we could extend the `license` store to do the same thing.

ThatGalNatalie:
For this, I don't think we should use a global mixin. Most of the articles that I read indicate that a global mixin should be used with caution as it will affect every Vue instance created afterwards and it could also cause naming conflicts. In addition, it will also affect third party components and I don't think every single component will need to know about license/permissions so we're  not really using it for its intended use that way. Thoughts? 

So far, I have only found 6 of the same license/permission checks (`flowConcurrency`, `isSelfServe`, `isUsageBased`, `isCurrent`, `planType`, and `hasRBAC`)

znicholasbrown:
That makes sense - is this an action on the license store then?

ThatGalNatalie:
I was thinking more of getters

zhen0:
Closed by #763

PR #763, opened 2021-04-12
licensing and permissions

ThatGalNatalie:
PR Checklist:

## Describe this PR
This will add three new getters to the license store:
planType 
hasPermissions 
allowedUsers 

#718


znicholasbrown (changes requested):
The permissions section of this is looking good but I'd like to remove the need for a specific plan name from the components; it'd be great if components could pass `FREE`, `STARTER`, `STANDARD`, and `ENTERPRISE` without the need for the year; this will let us change plans centrally instead of in every component.

zhen0:
I agree with Nicholas's comment @ThatGalNatalie - adding a check in the store which can tell is a license type is free, standard or starter, irrespective of the year (or any other details that might get added) will help cut down the amount of code we need to update in future.

zhen0 (dismissed):
Thanks for the changes @ThatGalNatalie - this looks good to me. Can you remove the package-lock.json change though?

zhen0 (dismissed):
Thanks @ThatGalNatalie LGTM!

znicholasbrown (approved):

PR #766, opened 2021-04-12
Add loading state to sign out button

znicholasbrown:
PR Checklist:

## Describe this PR
Adds a loading state to the sign out button, to make the flow feel more responsive.

### DIFF
--- a/src/components/Nav/UserMenu.vue
+++ b/src/components/Nav/UserMenu.vue
@@ -7,6 +7,7 @@ export default {
   data() {
     return {
       clockInterval: null,
+      loading: false,
       model: false,
       routes: [
         {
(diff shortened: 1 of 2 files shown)


src/components/Nav/UserMenu.vue
```diff
@@ -41,8 +42,14 @@ export default {
...
     async wipeClientAndLogout() {
-      await this.logout(true) // Pass true here to propagate this to all clients
-      document.querySelector('.router-view').style.opacity = 0
+      this.loading = true
+
+      try {
+        await this.logout(true) // Pass true here to propagate this to all clients
+        document.querySelector('.router-view').style.opacity = 0
+      } catch {
+        this.loading = false
```
whimsicallyson, line 51: do we want this outside of the catch, or are we pushing you off the page on logout so it doesn't matter?

znicholasbrown: We push off the page so we should only need to reset the loading state when there's an error

whimsicallyson (approved):

PR #768, opened 2021-04-12
kinda hacky fix for weird rounding bug

whimsicallyson:
PR Checklist:

## Describe this PR
Saw this weird display bug earlier today, so added in a display fix because javascript doesn't have a "round to certain precision"
Screen Shot 2021-04-12 at 1 39 25 PM

### DIFF
--- a/src/pages/Dashboard/UsageTiles/CycleUsage-Tile.vue
+++ b/src/pages/Dashboard/UsageTiles/CycleUsage-Tile.vue
@@ -28,7 +28,7 @@ export default {
     freeUsage() {
       if (isNaN(this.usage)) return null
       const percentage = this.usage / 10000
-      return percentage > 1 ? 100 : (percentage * 100).toFixed(2)
+      return percentage > 1 ? 100 : (percentage * 100).toFixed()
     },
     usageLoading() {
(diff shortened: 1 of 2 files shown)


src/pages/Dashboard/UsageTiles/CycleUsage-Tile.vue
```diff
@@ -28,7 +28,7 @@ export default {
     freeUsage() {
       if (isNaN(this.usage)) return null
       const percentage = this.usage / 10000
-      return percentage > 1 ? 100 : percentage * 100
+      return percentage > 1 ? 100 : (percentage * 100).toFixed(2)
```
znicholasbrown, line 31: I wonder if we need the decimal at all, maybe we can use `.toFixed()` instead

znicholasbrown (changes requested):

src/pages/Dashboard/UsageTiles/CycleUsage-Tile.vue
whimsicallyson: I'm fine with that!

znicholasbrown (approved):
LGTM 👍

PR #769, opened 2021-04-12
Add logging level to run config

zangell44:
PR Checklist:

This PR adds a drop down to the Run page, allowing users to specify logging level for a flow run.

I've never worked with Vue before, please be gentle :).

Open questions
- Testing. Based on the tests folder, it didn't seem like `Run.vue` was being tested? I assume I'm missing something here. If you can direct me to the right file, I'm happy to write the appropriate tests.

TODO
- [x] Confirm styling works well on both light and dark mode
- [x] Update tooltip with Core version once this PR is merged and released PrefectHQ/prefect#4383

### DIFF
--- a/src/pages/Flow/Run.vue
+++ b/src/pages/Flow/Run.vue
@@ -501,7 +501,17 @@ export default {
                 <template #item="{ item }">
                   <div>
                     <v-chip
-                      class="ma-2 debuglevel"
+                      class="ma-2 debuglevel cursor-pointer"
+                      :class="item.text.toLowerCase()"
+                    >
+                      {{ item.text }}
+                    </v-chip>
+                  </div>
+                </template>
+                <template #selection="{ item }">
+                  <div>
+                    <v-chip
+                      class="ma-2 debuglevel cursor-pointer"
                       :class="item.text.toLowerCase()"
                     >
                       {{ item.text }}
--- a/src/pages/Flow/Run.vue
+++ b/src/pages/Flow/Run.vue
@@ -828,9 +828,7 @@ export default {
     width: 100%;
   }
 }
-</style>
 
-<style lang="scss">
 .ma-2.debuglevel {
   color: var(--v-utilGrayLight-lighten5);
   font-weight: bold;
@@ -861,7 +859,9 @@ export default {
     background: var(--v-failRed-base);
   }
 }
+</style>
 
+<style lang="scss">
 .run-tab-icon {
   .svg-inline--fa {
     --fa-primary-color: var(--v-appForeground-base);
--- a/src/pages/Flow/Run.vue
+++ b/src/pages/Flow/Run.vue
@@ -830,7 +830,7 @@ export default {
 }
 
 .ma-2.debuglevel {
-  color: var(--v-utilGrayLight-lighten5);
+  color: var(--v-primaryLight-lighten5);
   font-weight: bold;
   justify-content: center;
   width: 75px;
--- a/src/pages/Flow/Run.vue
+++ b/src/pages/Flow/Run.vue
@@ -489,7 +489,7 @@ export default {
                     </p>
                     <p>
                       Please note this is only guaranteed to work for Agents
-                      running Prefect Core TODO or later.
+                      running Prefect Core 0.14.17 or later.
                     </p>
                   </MenuTooltip>
                 </div>
--- a/src/pages/Flow/Run.vue
+++ b/src/pages/Flow/Run.vue
@@ -42,6 +42,17 @@ export default {
     return {
       stickyActions: false,
 
+      // Logging Level
+      loggingLevel: 'DEFAULT',
+      loggingLevels: [
+        { text: 'Default', value: 'DEFAULT' },
+        { text: 'Debug', value: 'DEBUG' },
+        { text: 'Info', value: 'INFO' },
+        { text: 'Warn', value: 'WARN' },
+        { text: 'Error', value: 'ERROR' },
+        { text: 'Critical', value: 'CRITICAL' }
+      ],
+
       // Parameters
       parameters: null,
 
@@ -149,7 +160,9 @@ export default {
 
       try {
         this.loading = true
-
+        // if the user has specified a logging level, pass it
+        // to the run config as an env var
+        this.overrideLoggingEnvironmentVariable()
         const { data, errors } = await this.$apollo.mutate({
           mutation: require('@/graphql/Mutations/create-flow-run.gql'),
           variables: {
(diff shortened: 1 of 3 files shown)


src/pages/Flow/Run.vue
```diff
@@ -444,6 +465,53 @@ export default {
...
+                      This can also be controlled by providing an environment
+                      variable, "PREFECT__LOGGING__LEVEL".
+                    </p>
+                    <p>
+                      Any dropdown selection besides "Default" will override the
+                      environment variable.
+                    </p>
+                    <p>
+                      Please note this is only guaranteed to work for Agents
+                      running Prefect Core TODO or later.
```
whimsicallyson, line 492: definitely want to update this :)

src/pages/Flow/Run.vue
```diff
@@ -763,6 +831,37 @@ export default {
 </style>
 
 <style lang="scss">
+.ma-2.debuglevel {
```
whimsicallyson, line 834: in general we don't want to attach styles to built-in Vuetify class names, especially without specifying that the SCSS is scoped to the file (to do so, add "scoped" to line 833, but without digging I don't know if there's a reason we didn't do that on this file). Does this CSS work if you just specify `debuglevel` for the classname?

Also, did you check that all these color combos work in both light and dark modes?

src/pages/Flow/Run.vue
```diff
@@ -444,6 +465,53 @@ export default {
...
+                </div>
+              </div>
+            </v-col>
+
+            <v-col cols="12" md="9" class="mt-n4 mt-md-0">
+              <v-select v-model="loggingLevel" outlined :items="loggingLevels">
+                <template #item="{ item }">
+                  <div>
+                    <v-chip
+                      :class="['ma-2', 'debuglevel', item.text.toLowerCase()]"
```
whimsicallyson, line 504: pretty sure this would be more legible as 
```
class="ma-2 debuglevel"
:class="item.text.toLowerCase"
```

(I'm not 100% on my syntax there, might still need the parens on toLowerCase, so I'd test before taking this suggestion wholesale)

src/pages/Flow/Run.vue
whimsicallyson: Also, our linter has a strong preference (if not fully erroring) for kebab-case in classnames

zangell44: This CSS needs `ma-2` to work unfortunatley, I'll move this into the scoped style section of the file

zangell44: Changed the font color to look a bit better in dark mode, good catch!

src/pages/Flow/Run.vue
zangell44: That is very cool!

I'm having some trouble getting it to work though, is it worth setting aside some time to pair?

whimsicallyson: oh! I misunderstood what you were doing here. try `class="ma-2 debuglevel" :class="item.text.toLowerCase()"` (I read it initially as you were trying to make the words lowercase, not that you were adding a classname and making sure _that_ was cased properly. sorry!)

zangell44: Ahhh that makes sense now. Still cool!

Updated the syntax in my latest commit

whimsicallyson (dismissed):
With two caveats:
- I didn't have anything with interesting logging to test the output side of things
- We still need a version number to fill in from the core side

I'm happy with this!

src/pages/Flow/Run.vue
```diff
@@ -444,6 +465,54 @@ export default {
...
+                </div>
+              </div>
+            </v-col>
+
+            <v-col cols="12" md="9" class="mt-n4 mt-md-0">
+              <v-select v-model="loggingLevel" outlined :items="loggingLevels">
+                <template #item="{ item }">
+                  <div>
+                    <v-chip
+                      class="ma-2 debuglevel"
```
znicholasbrown, line 504:
```suggestion
                      class="ma-2 debuglevel cursor-pointer"
```
Right now the chip element has `cursor: default` on it, which makes it seem like a different interaction, this will fix that.

src/pages/Flow/Run.vue
```diff
@@ -444,6 +465,54 @@ export default {
...
+                      Please note this is only guaranteed to work for Agents
+                      running Prefect Core TODO or later.
+                    </p>
+                  </MenuTooltip>
+                </div>
+              </div>
+            </v-col>
+
+            <v-col cols="12" md="9" class="mt-n4 mt-md-0">
+              <v-select v-model="loggingLevel" outlined :items="loggingLevels">
```
znicholasbrown, line 500: I'd add another template override here for the selected item for style continuity, something like this:
```
<template #selection="{ item }">
  <v-chip class="ma-2 debuglevel cursor-pointer" :class="item.text.toLowerCase()">
    {{ item.text}}
  </v-chip>
</template>
```
I'm not going to add this as a suggestion cause the formatting is likely off

src/utils/runConfigs.js
```diff
@@ -27,7 +27,14 @@ export const UniversalRun = {
...
   icon: 'fad fa-globe',
-  args: []
+  args: [
+    {
+      arg: 'env',
+      input_type: 'object',
+      label: 'Environment variables',
+      description: 'Additional environment variables to set in the container.'
+    }
+  ]
```
znicholasbrown, line 37: I don't think this exists in the current version of `UniversalRun` but I might be wrong?

znicholasbrown (dismissed):
Looking pretty good to me @zangell44 - just a couple of comments. 

Note for the future: since we're adding run configs as a configurable default soon, setting the debug level will be a special case we'll want to parse and pull out, otherwise it'll be listed in the environment variables (which would be unexpected)

src/utils/runConfigs.js
zangell44: You are correct! But it will exist soon https://github.com/PrefectHQ/prefect/pull/4383/files#diff-5f6b81fc3cefe99964972a1b88d0f0769a472d896c38b4f0e6a9a07cdb399d42R46.

This PR can't be merged until the core side is all set.

src/pages/Flow/Run.vue
zangell44: this looks so clean! i was wondering if there was an easy way to do this

src/utils/runConfigs.js
znicholasbrown: Ahhhhh ok I thought I remembered that but wasn't sure if it was a dream 😅

src/pages/Flow/Run.vue
zangell44: resolved in latest commit

zangell44:
@znicholasbrown Core 0.14.17 is scheduled to be released today, this PR is ready to go as soon as it's released

zhen0 (approved):
This looks great @zangell44!

PR #770, opened 2021-04-13
update deploy tutorial for API keys

whimsicallyson:
PR Checklist:

## Describe this PR
@zhen0 noticed that the tutorial in the UI was still referencing tokens instead of API keys. Good catch!

### DIFF
--- a/src/pages/Tutorials/Markdown/Universal-Deploy.md
+++ b/src/pages/Tutorials/Markdown/Universal-Deploy.md
@@ -95,7 +95,7 @@ flow.register(project_name="tester")
 flow.run_agent()
 ```
 
-Click `"Copy"` above to copy the updated flow code. Paste the code into your interactive Python REPL session. If all goes well, you should see the local agent process start to run. If you're seeing the error message `"No agent API token provided"`, try passing in the agent token explicitly to the `run_agent()` method:
+Click `"Copy"` above to copy the updated flow code. Paste the code into your interactive Python REPL session. If all goes well, you should see the local agent process start to run. If you're seeing the error message `"No agent API token provided"`, try passing your API key explicitly to the `run_agent()` method:
 
 ```python
 flow.run_agent(token="<COPIED_KEY>")
--- a/src/pages/Tutorials/Markdown/Universal-Deploy.md
+++ b/src/pages/Tutorials/Markdown/Universal-Deploy.md
@@ -44,33 +44,33 @@ If you're running into issues, check that your Python environment is properly se
 
 # Log into Prefect Cloud from the CLI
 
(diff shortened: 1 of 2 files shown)


src/pages/Tutorials/Markdown/Universal-Deploy.md
```diff
@@ -44,33 +44,33 @@ If you're running into issues, check that your Python environment is properly se
...
 
-To authenticate, you'll need to create a [Personal Access Token](/user/tokens) and configure it with the [Prefect Command Line Interface](https://docs.prefect.io/orchestration/concepts/cli.html#cli).
+To authenticate, you'll need to create an [API Key](/user/keys) and configure it with the [Prefect Command Line Interface](https://docs.prefect.io/orchestration/concepts/cli.html#cli).
 
-- In the hamburger menu in the top left corner go to **User -> Personal Access Tokens -> Create A Token**.
-- Copy the created token
-- Configure the CLI to use the access token by running
+- In the user menu in the top right corner go to **Account Settings -> API Keys -> Create An API Key**.
+- Copy the created key
+- Configure the CLI to use the key by running
```
znicholasbrown, line 51: Can't we remove either the `prefect auth login` or creating a service account steps? I think these keys should be the same now, just scoped to different users.

znicholasbrown (changes requested):

src/pages/Tutorials/Markdown/Universal-Deploy.md
whimsicallyson: @znicholasbrown  We can remove the service account bit, and use the user's key to run the agent. I wasn't able to log in as a service account, which makes sense.
It seemed like we were considering service accounts a best practice for running agents though, is that not the case (alternatively, something we're not worried about for a getting started tutorial)?

whimsicallyson: chatted in slack, decided to go with just the user key for simplicity's sake

src/pages/Tutorials/Markdown/Universal-Deploy.md
```diff
@@ -110,7 +98,7 @@ flow.run_agent()
 Click `"Copy"` above to copy the updated flow code. Paste the code into your interactive Python REPL session. If all goes well, you should see the local agent process start to run. If you're seeing the error message `"No agent API token provided"`, try passing in the agent token explicitly to the `run_agent()` method:
```
znicholasbrown, line 98:
```suggestion
Click `"Copy"` above to copy the updated flow code. Paste the code into your interactive Python REPL session. If all goes well, you should see the local agent process start to run. If you're seeing the error message `"No agent API token provided"`, try passing in your API key explicitly to the `run_agent()` method:
```
I think this needs to be modified as well

znicholasbrown (approved):
Thanks for the changes @whimsicallyson

PR #771, opened 2021-04-13
Remove precommit hook

znicholasbrown:
PR Checklist:

## Describe this PR
- Removes the pre-commit hook
- Updates lint commands to use only those defined by `eslint` (which includes prettier rules)
- Adds a CI job for linting which uses the updated commands

### DIFF
--- a/src/components/ChartCard/ChartCardHeader-Ring.vue
+++ b/src/components/ChartCard/ChartCardHeader-Ring.vue
@@ -69,7 +69,7 @@ export default {
         :class="mainClasses"
       >
         {{ chartOverlayMain
-        }}<span class="title font-weight-black">{{ chartOverlayUnits }}</span>
+        }}<span class="text-h6 font-weight-black">{{ chartOverlayUnits }}</span>
       </div>
       <div
         class="font-weight-light"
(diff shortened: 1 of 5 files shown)


src/components/ChartCard/ChartCardHeader-Ring.vue
```diff
@@ -115,7 +115,7 @@ export default {
 
       <div class="text-h3 font-weight-bold d-flex align-start justify-end">
         {{ chartOverlayMain
-        }}<span class="title font-weight-black">{{ chartOverlayUnits }}</span>
```
zangell44, line 118: For my own knowledge, how would I know if I'm breaking something here? My concern was we might be using styling in which `title` had expected styling which `text-h6` would not.

zangell44 (approved):
love it

src/components/ChartCard/ChartCardHeader-Ring.vue
znicholasbrown: Great question! If you're running the development server, deprecated classes like this will get caught by the prettier rules coming from Vuetify; you'd see an overlay on the application that tells you which classes are incorrect/deprecated. In this case, these components are part of our library but aren't being used anywhere at the moment and so don't get parsed when the dev server is running (hence these didn't get caught when I updated all the classes before). If you were to import either of these components this would have shown up immediately.

whimsicallyson: this was from an overarching change in how Vuetify names their text classes, Nicholas did a sweep of class names a week or so ago but may have just missed this one.

Issue #614, opened 2021-02-16
Provide a better way to get back to the Server application


znicholasbrown:
**Why**

If a user clicks the `Cloud` or `Server` buttons from the connection menu, the application sets a `localStorage` variable, `backend`, which remembers the selection and brings them through the associated startup logic. If Cloud is selected, the only way for users to get back to the Server application is to log in to a Cloud account (or create one if they don't already have one) and then select Server again. 

**What**

* [ ] a user can easily get back to the Server application without a Cloud account

**Notes**

PR #772, opened 2021-04-13
don't save backend in localstorage

whimsicallyson:
PR Checklist:

## Describe this PR
I believe this is all we should need to solve #614. By not saving the backend choice in localstorage (relying instead on the `VUE_APP_BACKEND` var), if a user without cloud switches the toggle they won't be forever stuck in a redirect to universal.prefect.io


znicholasbrown (approved):
LGTM 👍

znicholasbrown:
Resolves: #614

Issue #765, opened 2021-04-12
BUG: Can't create schedule if default parameters are not provided


mashun4ek:
**Bug Description**

**Steps To Reproduce**

1. Use following flow:
```
from prefect import task, Flow, Parameter
import prefect
@task(log_stdout=True)
def say_hello(name):
    print("Hello, {}!".format(name))
name = Parameter("name", required=False)
with Flow("Test not required parameters with schedule") as flow:
    say_hello(name)
flow.register(project_name="Training")
```
2. Create a schedule via UI without parameters.
3. Following notification will appear: 
`There was a problem creating your schedule, please try again shortly. Error message: Error: GraphQL error: Invalid clock provided for schedule: {'type': 'IntervalClock', 'interval': 60000000, 'parameter_defaults': {'name': None}, 'start_date': {'dt': '2021-04-12T12:50:37.712905', 'tz': 'America/New_York'}}`.

**Browsers Tested:**
* [x] Chrome
* [ ] Firefox
* [ ] Safari
* [ ] Edge
* [ ] IE

----------
**Notes**

Slack: https://prefect-community.slack.com/archives/C0192RWGJQH/p1617103607042800

zhen0:
This should be addressed by https://github.com/PrefectHQ/ui/pull/773

PR #773, opened 2021-04-13
default parameters 

ThatGalNatalie:
PR Checklist:

## Describe this PR
#765

### DIFF
--- a/src/components/CustomInputs/DictInput.vue
+++ b/src/components/CustomInputs/DictInput.vue
@@ -52,7 +52,7 @@ export default {
       jsonInput: '{}',
       keys: [],
       values: [],
-      paramsToAdd: []
+      includedKeys: []
     }
   },
   computed: {
@@ -65,7 +65,7 @@ export default {
     value() {
       const dict = {}
       this.keys.forEach((k, i) => {
-        if (k && (!this.includeCheckbox || this.paramsToAdd.includes(k))) {
+        if (k && (!this.includeCheckbox || this.includedKeys.includes(k))) {
           dict[k] = this.values[i]
         }
       })
@@ -89,7 +89,7 @@ export default {
         })
       }
     },
-    paramsToAdd() {
+    includedKeys() {
       this.$emit('input', { ...this.value })
       this.jsonInput = this.keys.length > 0 ? JSON.stringify(this.value) : '{}'
     }
@@ -214,7 +214,7 @@ export default {
         >
           <v-col v-if="includeCheckbox" cols="1">
             <v-checkbox
-              v-model="paramsToAdd"
+              v-model="includedKeys"
               multiple
               :value="keys[i]"
             ></v-checkbox>
--- a/src/components/CustomInputs/DictInput.vue
+++ b/src/components/CustomInputs/DictInput.vue
@@ -243,7 +243,9 @@ export default {
                   ? dict[i].value.toString()
                   : valueLabel
               "
-              :readonly="!includedKeys.includes(keys[i])"
+              :readonly="
+                includeCheckbox ? !includedKeys.includes(keys[i]) : false
+              "
               @keyup="_handleKeypress"
             />
           </v-col>
--- a/src/components/CustomInputs/DictInput.vue
+++ b/src/components/CustomInputs/DictInput.vue
@@ -43,6 +43,11 @@ export default {
       type: String,
       requird: false,
       default: 'Value'
+    },
+    defaultCheckedKeys: {
+      type: Array,
+      required: false,
+      default: () => []
     }
   },
   data() {
@@ -52,7 +57,7 @@ export default {
       jsonInput: '{}',
       keys: [],
       values: [],
-      includedKeys: []
+      includedKeys: this.defaultCheckedKeys
     }
   },
   computed: {
--- a/src/components/CustomInputs/DictInput.vue
+++ b/src/components/CustomInputs/DictInput.vue
@@ -64,9 +64,10 @@ export default {
     },
     value() {
       const dict = {}
-      this.keys
-        .filter(k => k !== null && this.paramsToAdd.includes(k))
-        .map((k, i) => (dict[k] = this.values[i]))
+      const keys = this.includeCheckbox
+        ? this.keys.filter(k => this.paramsToAdd.includes(k))
+        : this.keys
+      keys.filter(k => k !== null).map((k, i) => (dict[k] = this.values[i]))
       return dict
     }
   },
@@ -86,6 +87,9 @@ export default {
           this.$refs['json-input'].validateJson()
         })
       }
+    },
+    paramsToAdd() {
+      this.$emit('input', { ...this.value })
     }
   },
   mounted() {
--- a/src/components/CustomInputs/DictInput.vue
+++ b/src/components/CustomInputs/DictInput.vue
@@ -63,12 +63,25 @@ export default {
       return this.keys.map((p, i) => i)
     },
     value() {
-      const dict = {}
-      const keys = this.includeCheckbox
-        ? this.keys.filter(k => this.paramsToAdd.includes(k))
-        : this.keys
-      keys.filter(k => k !== null).map((k, i) => (dict[k] = this.values[i]))
-      return dict
+      // const dict = {}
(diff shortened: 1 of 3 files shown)


src/components/CustomInputs/DictInput.vue
```diff
@@ -58,9 +64,10 @@ export default {
     },
     value() {
       const dict = {}
-      this.keys
-        .filter(k => k !== null)
-        .map((k, i) => (dict[k] = this.values[i]))
+      const keys = this.includeCheckbox
+        ? this.keys.filter(k => this.paramsToAdd.includes(k))
+        : this.keys
+      keys.filter(k => k !== null).map((k, i) => (dict[k] = this.values[i]))
```
zhen0, line 70: Index here doesn't match up with the correct value.  If I only add one param it takes the value from position 0 in the values array.

zhen0 (changes requested):

src/pages/Flow/Settings/Schedules.vue
```diff
@@ -290,10 +321,22 @@ export default {
                   />
                 </div>
                 <div v-show="selectedTab === 1">
+                  <p
+                    v-if="checkDefualtParameters(allDefaultParameters)"
+                    class="mt-8 text-body-1"
+                    >If a clock has default parameters then that will override
+                    the flow group and flow parameters. If not, the flow group
+                    paramter will be used, else it will use the flow parameter.
```
madkinsz, line 329:
```suggestion
                    parameter will be used, else it will use the flow parameter.
```

src/pages/Flow/Settings/Schedules.vue
```diff
@@ -290,10 +321,22 @@ export default {
                   />
                 </div>
                 <div v-show="selectedTab === 1">
+                  <p
+                    v-if="checkDefualtParameters(allDefaultParameters)"
+                    class="mt-8 text-body-1"
+                    >If a clock has default parameters then that will override
```
madkinsz, line 327:
```suggestion
                    >If a clock has default parameters then they will override
```

src/pages/Flow/Settings/Schedules.vue
madkinsz: Probably should not just be ", else". Perhaps it should just say something like

Parameters will be resolved in the following order: clock values, flow group values, flow values. If a parameter value is set in an earlier item, it will not be loaded from a later one.

Also shouldn't flow settings have precedence over flow group settings? i.e. clock > flow > flow group

zhen0: Thanks - I've included some of that wording in my update.

src/components/CustomInputs/DictInput.vue
```diff
@@ -57,11 +63,23 @@ export default {
...
+      for (const [key, value] of Object.entries(obj)) {
+        if (this.includeCheckbox) {
+          if (this.paramsToAdd.includes(key)) {
+            parametersDict.push({ [key]: value })
+          }
+        } else {
+          parametersDict.push({ [key]: value })
+        }
+      }
+      return Object.assign({}, ...parametersDict)
```
znicholasbrown, line 82: This isn't the correct place for this or at least is incorrectly and confusingly named for what the component is meant to do; this block, and the `paramsToAdd` prop, are specific to the parameters implementation only and mean that an otherwise generic component/interface has to know something about its parent. This is also a really confusing block, couldn't we write it like this?

```suggestion
      const dict = {}
      this.keys
        .filter(k => k !== null)
        .filter(k => !this.includeCheckbox || this.paramsToAdd.includes(k))
        .map((k, i) => (dict[k] = this.values[i]))
      return dict
```

znicholasbrown (changes requested):
Overall this works but I have a few comments:
1. Generic components shouldn't include variables and methods pertaining to a specific parent implementation (in this case, anything to do with parameters)
2. Right now, values are opt-in, rather than opt-out; I think it should be the opposite, given that it's unclear what the checkbox does without reading the block of text at the top
3. Unchecked kv pairs should be disabled

src/pages/Flow/Settings/Schedules.vue
```diff
@@ -290,10 +321,20 @@ export default {
...
                 <div v-show="selectedTab === 1">
+                  <p
+                    v-if="checkDefualtParameters(allDefaultParameters)"
+                    class="mt-8 text-body-1"
+                    >If checked, these parameters will be included on your
+                    schedule. If a parameter value is included here, it will not
+                    be updated by later changes to your flow or flow group
+                    parameters.
+                  </p>
+
```
znicholasbrown, line 332:
```suggestion
                  <p
                    v-if="checkDefualtParameters(allDefaultParameters)"
                    class="mt-8 text-body-1"
                    >
                      Checked parameters will override their corresponding defaults for runs generated from this schedule.
                  </p>

```

src/components/CustomInputs/DictInput.vue
ThatGalNatalie: For the suggested code,  if you have two or more parameters and then select only one parameter then the index will be at the 0 position and it will give the wrong value which is what the code above was trying to change. I agree that I think it could be written better

ThatGalNatalie: We also included a `includeCheckbox`prop which I think that would give the component the ability to opt out?

znicholasbrown: Oops good point, you could do this instead:
```
      const dict = {}
      this.keys.forEach((k, i) => {
        if (k && (!this.includeCheckbox || this.paramsToAdd.includes(k))) {
          dict[k] = this.values[i]
        }
      })
      return dict
```

znicholasbrown: To your second point, yes, the prop will allow an instantiation to ignore that logic but the point is that the logic should be generic anyway, i.e. shouldn't contain references to a specific data instance. The component should be able to be read as completely agnostic to the business logic but using variables and methods that are implementation-specific (even if they're opt-out) hinders that readability. Params !== keys and mean the component can't be taken in isolation without understanding what parameters are. In this case the swap is pretty easy, since changing the variable/method names is probably all that's required. So instead of `paramsToInclude`, something like `includedKeys` is agnostic to any external data structure.

ThatGalNatalie: ahhh okay that makes sense

src/components/CustomInputs/DictInput.vue
```diff
@@ -224,6 +243,7 @@ export default {
                   ? dict[i].value.toString()
                   : valueLabel
               "
+              :readonly="!includedKeys.includes(keys[i])"
```
znicholasbrown, line 246: This should only be disabled if the `includeCheckbox` prop is passed as well, right?

znicholasbrown (changes requested):

src/components/CustomInputs/DictInput.vue
ThatGalNatalie: oooh my bad. I will add that now

ThatGalNatalie:
If you create a new schedule and you have  parameters but you don't check them, should all the default parameters show up when you go to modify that schedule or should it just say that it has no default parameters? 
@znicholasbrown

src/components/CustomInputs/DictInput.vue
```diff
@@ -224,6 +248,9 @@ export default {
                   ? dict[i].value.toString()
                   : valueLabel
               "
+              :readonly="
```
zhen0, line 251: Is there a reason this is readonly instead of disabled?  My preference would be to use disabled here as it's clearer that it's deliberately disabled.  Read-only feels odd as I can focus on it but not change the value.

znicholasbrown: This was primarily to match styling of the key field, but I'm not sure the distinction between readonly and disabled is particularly important when the field is disabled either way

znicholasbrown:
@ThatGalNatalie - all the defaults should show up, unchecked

znicholasbrown:
Thanks for the changes @ThatGalNatalie - this is looking good!

znicholasbrown (approved):

PR #774, opened 2021-04-14
Fix some issues with checking auth when returning to the application

znicholasbrown:
PR Checklist:

## Describe this PR
This PR fixes the ordering of some auth checks when returning to the application and attempts to address some issues where tenants weren't being fetched after the apollo cache had been cleared and no navigation had taken place.

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -191,6 +191,7 @@ export default {
     isAuthorized(value) {
       if (value) {
         this.getApi()
+        this.getTenants()
       }
     }
   },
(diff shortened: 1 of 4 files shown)


zhen0:
@znicholasbrown  - I can't reproduce reliably but I'm seeing odd behavior where flows from the prefect tenant show in my personal tenant in dev.  You can see examples from this morning from my user in log rocket.  I've pulled this branch locally to see if I can reproduce in a more useful way but I don't think we can approve until we're sure it's not connected to this PR. (And actually we should find out the cause either way!)

zhen0 (approved):
Thanks @znicholasbrown

PR #783, opened 2021-04-20
Fix auth loop

znicholasbrown:
PR Checklist:

## Describe this PR
Looks to fix intermittent auth loops by removing auth and tenant nav guards and instead providing a `setup` method that runs before application rendering and performs auth, fetches tenants, and sets the current tenant.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -8,7 +8,7 @@
 
 ### Bugfixes
 
-- None
+- Fix auth loop - [#783](https://github.com/PrefectHQ/ui/pull/783)
 
 ## 2021-04-15
 
--- a/src/main.js
+++ b/src/main.js
@@ -191,8 +191,13 @@ if (TokenWorker?.port) {
     const type = e.data?.type
     const payload = e.data?.payload
 
-    // eslint-disable-next-line no-console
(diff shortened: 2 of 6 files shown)


whimsicallyson (approved):
code looks good! so does testing!

wrightjonathan:
I haven't looked into the details yet, but the latest UI release which I believe only includes changes from this ticket has prevented our users accessing an instance of Prefect Server hosted on our infrastructure. I've rolled back to image ui:2021-04-15, which we were previously using, and it is working again. It could be that it is entirely related to our configuration, but I wanted to report it in case it affects anyone else or someone immediately knows what our issue might be.

Here's what was happening. Very soon after the UI starts to load in the browser we were being redirected to: https://universal.prefect.io/oauth2/  ... omitted rest of the url ...
Showing this message.
Screenshot 2021-04-22 at 11 30 00
In our case 'redirect_uri' is set to the domain name of our Prefect server.

madkinsz:
Hey @wrightjonathan I think this is being addressed at #788

imranq2:
We're seeing the exact same error after upgrading to 2021-04-22

PR #787, opened 2021-04-22
allow editing in json editor for checked params

ThatGalNatalie:
PR Checklist:

## Describe this PR
The JSON editor wasn't reflecting the checked/unchecked parameters. When modifying an existing schedule, the checked parameters should now be the only parameters that is displayed


ThatGalNatalie:
@dylanbhughes

dylanbhughes:
Having trouble looking in Netlify but the code makes sense to me!

zhen0:
This addresses the issue it was asked to nicely but I fear that the JSON edit experience could still lead to mistakes.  If I go to the schedule parameters tab and immediately switch into JSON, add my parameters and then switch back to set the schedule, I'd miss that I need to check the box to include my parameters.  Could we add a JSON edit specific warning?

zhen0 (changes requested):
I think this fixes the issue @ThatGalNatalie - thanks.  However, I think we risk wiping all a user's JSON input.  Can we add a warning that they also need to check the boxes?  Or remove the need to check boxes?

znicholasbrown:
Previews enabled for this PR 👍

src/components/CustomInputs/DictInput.vue
```diff
@@ -92,11 +95,13 @@ export default {
         this.$nextTick(() => {
           this.$refs['json-input'].validateJson()
         })
+      } else {
+        this.includedKeys = Object.keys(this.value)
       }
     },
-    includedKeys() {
+    includedKeys(val) {
+      this.jsonInput = val.length > 0 ? JSON.stringify(this.value) : '{}'
```
dylanbhughes, line 103: Makes sense!

ThatGalNatalie:
update:
Latest commit should allow you to edit in the JSON editor and the parameters should be checked/unchecked based on that. 

However, if you check a parameter then it will be added to the jsoninput but if you decide to uncheck a parameter it will not be reflected in the json editor. I think it has something to do with the `this.value` computed prop and how we’re checking what gets returned but can’t seem to wrap my head around it and need some help on that. 

Tested the run page briefly and it seems to not be affected by this change but will need to do a more thorough test

src/components/CustomInputs/DictInput.vue
```diff
@@ -92,11 +95,13 @@ export default {
         this.$nextTick(() => {
           this.$refs['json-input'].validateJson()
         })
+      } else {
+        this.includedKeys = Object.keys(this.value)
```
znicholasbrown, line 99:
```suggestion
        try {
          this.includedKeys = Object.keys(JSON.parse(this.jsonInput))
        } catch {
          this.includedKeys = Object.keys(this.value)
        }
```

src/components/CustomInputs/DictInput.vue
```diff
@@ -70,7 +70,10 @@ export default {
     value() {
       const dict = {}
       this.keys.forEach((k, i) => {
-        if (k && (!this.includeCheckbox || this.includedKeys.includes(k))) {
+        if (
+          (k && (!this.includeCheckbox || this.includedKeys.includes(k))) ||
+          Object.keys(JSON.parse(this.jsonInput)).includes(k)
+        ) {
           dict[k] = this.values[i]
         }
```
znicholasbrown, line 78:
```suggestion
        if (
          (k && (!this.includeCheckbox || this.includedKeys.includes(k))) ||
          (this.json && Object.keys(JSON.parse(this.jsonInput)).includes(k))
        ) {
          dict[k] = this.values[i]
        }
```

znicholasbrown (changes requested):
2 suggestions which I think should fix the issue you're seeing

src/components/CustomInputs/DictInput.vue
ThatGalNatalie: ahhhh it was checking if they were on the json editor then push the keys/values but if not push the keys/values from the checkboxes

zhen0 (commented):
One small suggestion

src/components/CustomInputs/DictInput.vue
```diff
@@ -108,10 +113,23 @@ export default {
...
-        )
+        if (this.includeCheckbox) {
+          Object.keys(json).forEach(k => {
+            let i = this.keys.indexOf(k)
+
+            if (i > -1) {
+              this.values[i] = json[k]
+            } else {
+              this.keys.push(k)
+              this.values.push(json[k])
```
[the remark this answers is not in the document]
ThatGalNatalie: just pushed a new commit for this

znicholasbrown (approved):
This LGTM @ThatGalNatalie 👍

PR #802, opened 2021-04-29
Gracefully handle null default tenants

cicdw:
PR Checklist:

## Describe this PR
This PR gracefully handles the situation where an API key has no default tenant associated with it; at this moment in time, this isn't truly possible but it may be in the near future.  Without these changes, the API Keys screen fails to load whenever the user has keys with no default tenants.


zhen0 (approved):
LGTM @cicdw

PR #804, opened 2021-04-29
KV Store 

ThatGalNatalie:
PR Checklist:

## Describe this PR
Allow users to store small pieces of data relevant to their workflows

Screen Shot 2021-05-27 at 4 46 28 PM


src/pages/TeamSettings/KV.vue
```diff
@@ -0,0 +1,636 @@
...
+    }
+  },
+  watch: {
+    tenant() {
+      this.$apollo.queries.kv.refetch()
+    }
+  },
+  methods: {
+    checkValue(value) {
+      return typeof value == 'object' ? JSON.stringify(value) : value
```
zhen0, line 114: I get an error if I enter a number as a value. CodeMirror expects it to be input as a string.  Is there a reason we're only stringify-ing an object?

src/pages/TeamSettings/KV.vue
```diff
@@ -0,0 +1,636 @@
...
+                  @click="handleReset(item)"
+                >
+                  Reset
+                  <v-icon small>refresh</v-icon>
+                </v-btn>
+              </div>
+
+              <JsonInput
+                ref="kvRef"
+                v-model="KvValueInput"
```
zhen0, line 418: Add a prepend icon

src/pages/TeamSettings/KV.vue
```diff
@@ -0,0 +1,636 @@
...
+                      >Type</v-btn
+                    >
+                  </template>
+                  <v-list>
+                    <v-list-item-group
+                      v-model="selectedTypeIndex"
+                      color="primary"
+                    >
+                      <v-list-item
+                        v-for="(item, index) in kvTypes"
```
zhen0, line 475: I would switch item to type so that it's got a unique name.
```suggestion
                        v-for="(type, index) in kvTypes"
```

src/pages/TeamSettings/KV.vue
```diff
@@ -0,0 +1,636 @@
...
+                        :key="index"
+                      >
+                        <v-list-item-title>{{ item.text }} </v-list-item-title>
```
zhen0, line 478:
```suggestion
                        <v-list-item-title>{{ type.text }} </v-list-item-title>
```

zhen0 (dismissed):
Lots of small changes @ThatGalNatalie. Let me know if you have any questions. 

Also - please check if you want to include the package-lock.json

src/components/Nav/TeamMenu.vue
```diff
@@ -158,6 +158,22 @@ export default {
...
 
+        <v-list-item :disabled="!isCloud" :to="'/team/kv'">
+          <v-list-item-avatar tile>
+            <v-icon large color="primaryDark">
+              store
+            </v-icon>
+          </v-list-item-avatar>
+          <v-list-item-content>
+            <v-list-item-title class="text-subtitle-1">
+              KV
```
zhen0, line 169:
```suggestion
              KV Store
```

src/pages/TeamSettings/KV.vue
```diff
@@ -0,0 +1,669 @@
...
+</script>
+
+<template>
+  <div>
+    <ManagementLayout :show="!isFetchingKV" control-show>
+      <template #title>KV Store</template>
+
+      <template #subtitle>
+        <!-- Manage your
+        <ExternalLink href="https://docs.prefect.io/">key/value</ExternalLink> -->
```
zhen0, line 347:
```suggestion
        <ExternalLink href="https://docs.prefect.io/">key:value</ExternalLink> -->
```

zhen0 (commented):
Two small suggestions/nitpicks from me.  

Also could we add a default sort e.g. by Key - I found the keys jumping around when I add a new one or edit one.  I think you can do that with the sort-by prop?

zhen0 (commented):
This looks good to me but I'll let @znicholasbrown review before we merge.

znicholasbrown (dismissed):
First thoughts:
- No error message is thrown when the `set_key_value` mutation fails; we need to show a helpful message in particular for instances where key count is exceeded but in general the loading state shouldn't persist if the mutation fails
- Searchbar styling should match the searchbars on other team management pages:
Screen Shot 2021-05-24 at 1 39 35 PM
- The team menu icon should use an FA-style icon so it matches the rest of the non-Prefect icons; [this one](https://fontawesome.com/icons/brackets-curly?style=duotone) or [this one](https://fontawesome.com/icons/database?style=duotone) might be good candidates

Still need to do some testing so will add another review later 👍🏻

zhen0:
@znicholasbrown - This one is ready for your review also.

znicholasbrown (dismissed):
Some more thoughts:
- It'd be good to have a count of the current number of keys out of how many the license allows, and then to disable the `Add KV` button when that count is at max
- I can't currently set a value to null, I'd expect to be able to do this unless there's an API limitation to this (this applies to both creating a new key and updating a key value)
- It doesn't look like the KV table is polling at all, I need to reload the page to get updated key values; some generous polling would be good here, maybe 30 seconds?
- I'm not sure the `Created by` column is adding much value, since that only reflects the current team and switching a team loads only the values for that team
- I don't think this info box is needed now that the panels are expandable (it's also broken for long pieces of data):
Screen Shot 2021-05-27 at 3 10 07 PM

- I don't think we need the placeholder text of the reset value like here (this confused me when I tried to clear the box):
Screen Shot 2021-05-27 at 3 11 09 PM


Some small notes on the create modal:
Screen Shot 2021-05-27 at 3 20 15 PM
1. This header can be "Create KV"
2. Can just be "Key"
3. Can just be "Value" ( no need for the type validation hint either )

I'd also switch text of the "Confirm" button to "Save"

znicholasbrown (dismissed):
Thanks for the changes @ThatGalNatalie ! I've added a few comments re: loading states/race conditions. This is on a license without access to the feature:
Screen Shot 2021-06-01 at 10 19 46 AM

In a case like this I think it'd make the most sense to hide the table/add buttons and show something about contacting us to get access to the feature.

src/pages/TeamSettings/KV.vue
```diff
@@ -0,0 +1,692 @@
...
+      this.isSettingKV = true
+
+      if (this.selectedTypeIndex === 2 && !this.validKVJSON()) {
+        this.isSettingKV = false
+        this.invalidKV = true
+        return
+      }
+
+      if (this.isKvUpdate) {
+        this.deleteKV({ id: this.selectedKV?.id }, { isModifying: true })
```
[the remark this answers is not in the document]
ThatGalNatalie: This line is necessary when updating but not when creating. Will add that in.

src/pages/TeamSettings/KV.vue
```diff
@@ -0,0 +1,692 @@
...
+      },
+      fetchPolicy: 'network-only'
+    }
+  }
+}
+</script>
+
+<template>
+  <div>
+    <ManagementLayout :show="!isFetchingKV" control-show>
```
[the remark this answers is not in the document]
ThatGalNatalie: There is not a reason to not show that so I'll rm that

zhen0:
@znicholasbrown - we'd love to get this merged today so that Zach's blog post can go out. Let me know if there's anything I can do to help speed up review.

znicholasbrown (approved):
LGTM, thanks for the changes @ThatGalNatalie 👍🏻

PR #808, opened 2021-05-01
Created by field service accounts

zangell44:
PR Checklist:

- ~[ ] add/update tests (or don't, for reasons explained below)~ 

## Describe this PR
This PR adds "created by" information to the list of api keys associated with a service account.

### DIFF
--- a/src/graphql/Tokens/api-keys.js
+++ b/src/graphql/Tokens/api-keys.js
@@ -1,0 +1,33 @@
+import gql from 'graphql-tag'
+
+export default function(isCloud) {
+  return gql`
+    query ApiKeys(
+      $limit: Int
+    ) {
+    auth_api_key(
+        limit: $limit
+        order_by: { created: desc }
+      ) {
+
+        id
+        name
+        created
+        expires_at
+        default_tenant_id
+        user_id
+        updated
+
+      ${
+        isCloud
+          ? `
+          created_by {
+              id
+              username
+          }`
+          : ''
+      }
+    }
+    }
+`
+}
--- a/src/pages/TeamSettings/Service-Accounts-Table.vue
+++ b/src/pages/TeamSettings/Service-Accounts-Table.vue
@@ -237,8 +237,7 @@ export default {
     },
     keys: {
       query() {
-        // TODO - why isn't this.isCloud true for local dev?
-        return require('@/graphql/Tokens/api-keys.js').default(true)
+        return require('@/graphql/Tokens/api-keys.js').default(this.isCloud)
       },
       fetchPolicy: 'network-only',
       error() {
@@ -248,7 +247,6 @@ export default {
         )
       },
       result({ data }) {
-        console.log(data)
         this.keys = data.auth_api_key.map(key => {
           return {
             id: key.id,
@@ -256,7 +254,7 @@ export default {
             created_at: key.created,
             expires: key.expires_at,
             user_id: key.user_id,
-            created_by: key.created_by.username
+            created_by: this.isCloud ? key.created_by.username : ''
           }
         })
         this.isFetchingKeys = false
--- a/src/pages/TeamSettings/Service-Accounts-Table.vue
+++ b/src/pages/TeamSettings/Service-Accounts-Table.vue
@@ -236,7 +236,10 @@ export default {
       }
     },
     keys: {
-      query: require('@/graphql/Tokens/api-keys.gql'),
+      query() {
+        // TODO - why isn't this.isCloud true for local dev?
+        return require('@/graphql/Tokens/api-keys.js').default(true)
+      },
       fetchPolicy: 'network-only',
       error() {
         this.handleAlert(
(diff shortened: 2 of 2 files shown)


src/graphql/Tokens/api-keys.js
```diff
@@ -0,0 +1,33 @@
...
+        name
+        created
+        expires_at
+        default_tenant_id
+        user_id
+        updated
+
+      ${
+        isCloud
+          ? `
```
zhen0, line 23: I don't think we need the isCloud check here as this query will only be run by cloud.

src/pages/TeamSettings/Service-Accounts-Table.vue
```diff
@@ -236,7 +236,9 @@ export default {
       }
     },
     keys: {
-      query: require('@/graphql/Tokens/api-keys.gql'),
+      query() {
+        return require('@/graphql/Tokens/api-keys.js').default(this.isCloud)
```
zhen0, line 240: Again this page is only accessed by cloud so I think we can simplify this and run it as a standard query.

src/pages/TeamSettings/Service-Accounts-Table.vue
```diff
@@ -251,7 +253,8 @@ export default {
             name: key.name,
             created_at: key.created,
             expires: key.expires_at,
-            user_id: key.user_id
+            user_id: key.user_id,
+            created_by: this.isCloud ? key.created_by.username : ''
```
zhen0, line 257: Again, this should always be cloud.

src/graphql/Tokens/api-keys.js
zhen0: I'd revert to using the .gql query but add ```created_by {id, username}```

src/pages/TeamSettings/Service-Accounts-Table.vue
zhen0: Again we can be confident that this page is all cloud so can simply remove the check for isCloud. 

**For info** - Actually here you'd need to make sure you import the isCloud getter - without it this is returning false and so not showing the created_by when it is present.   the 'isCloud' getter looks like this ```...mapGetters('api', ['isCloud'])``` in the 'computed' section of a component and checks the api vuex store.

zhen0 (changes requested):
Thanks for this @zangell44 - a nice extra piece of information to have included.  My main comment is that the tokens page is only used in Cloud so you can remove the isCloud checks which should make sure usernames show up and simplify your queries!

zangell44:
Dang it, I was excited I got this to work haha. But yep fair point that is 10x easier if we don't need the `isCloud` checks, I'll update the branch accordingly

src/pages/TeamSettings/Service-Accounts-Table.vue
```diff
@@ -306,11 +307,9 @@ export default {
             >
               <v-list-item-title>{{ key.name }}</v-list-item-title>
               <v-list-item-subtitle
-                >Created
-                {{
-                  key.created_at ? formDate(key.created_at) : ''
-                }}</v-list-item-subtitle
-              >
+                >Created {{ key.created_at ? formDate(key.created_at) : '' }}
+                {{ key.created_by ? ' by '.concat(key.created_by) : '' }}
```
zhen0, line 311: This works but might be simpler/neater code to use a span and v-if
```suggestion
                <span v-if="key.created_by">by {{ key.created_by }}</span>
```

zhen0 (approved):
Looks good to me @zangell44 - one small suggestion that you can include or note for future vue use!

zhen0:
@zangell44 - can you update master again?  As this updates the graphql this will need approval from Chris or Nicholas too.  (Can @ them once master is updated.)

zangell44:
Merged latest master, cc @znicholasbrown @cicdw for graphql approval

src/graphql/Tokens/api-keys.gql
```diff
@@ -7,5 +7,10 @@ query APIKeys {
     default_tenant_id
     user_id
     updated
+
+    created_by {
+      id
+      username
+    }
```
znicholasbrown, line 14: Running this across a few tenants, I don't have a single key with this field populated. Is that expected?

znicholasbrown (approved):
One question, otherwise this lgtm

src/graphql/Tokens/api-keys.gql
zangell44: Thank you for testing! Yes! A series of Cloud PRs last week added this relation. This cloud PR needs to be released for the UI to work correctly PrefectHQ/cloud#3397

zangell44: *Added the relation + actually tracking who created a service account key

znicholasbrown: Ahh I missed that while I was out, thanks for the info! This is fine to merge then I think 👍

PR #811, opened 2021-05-05
Separate auth from the primary application

znicholasbrown:
PR Checklist:

## Describe this PR
This PR removes auth and token logic from the view layer to improve performance and maintainability. It also does the following:
- Removes backwards support for Auth0
- Centralizes support for browsers / environments without Service worker support
- Modularizes login/logout/refresh methods for use in both webkit and non-webkit browsers
- Removes Okta default HTML5 storage in favor of closure-first storage (with fallbacks)
- Fetches central data (including tenants, licenses, and users) before the view layer is instantiated
- Simplifies the Vue Apollo client middleware logic (should be much faster)
- Creates a global Apollo client constructor for use by services other than the view layer
- Removes all actions from the auth store
- Removes calls to Okta's `getUser` in favor of id token claims

TODO:
- [x] write tests the new auth layer
- [x] update existing tests
- [x] implement refresh logic for webkit browsers
- [x] watch id token refreshes, update subscribers accordingly
- [x] handle tenant switching with multiple tabs properly
- [x] remove deprecated auth logic

Known issues:
- [ ] tenant switching doesn't clear the apollo cache in Safari


src/app.js
```diff
@@ -0,0 +1,285 @@
...
+          try {
+            !blockedProps.includes(key) &&
+              !(key in this[key]?.$props) &&
+              (this[key] = null)
+          } catch {
+            /* */
+          }
+        })
+      } catch (e) {
+        // eslint-disable-next-line no-console
```
zhen0, line 243: Do we need to log this in all environments?

znicholasbrown: Hm probably not, this is carried over from the existing garbage collector.

babel.config.js
```diff
@@ -1,4 +1,7 @@
 module.exports = {
   presets: [['@vue/app', { useBuiltIns: 'entry' }]],
-  plugins: ['@babel/plugin-proposal-nullish-coalescing-operator']
+  plugins: [
+    '@babel/plugin-proposal-nullish-coalescing-operator',
+    '@babel/plugin-proposal-optional-chaining'
```
znicholasbrown, line 5: This was added back in to fix an issue parsing an upgrade of one of the graphql packages (`graphql-language-service`)

src/apollo.js
```diff
@@ -0,0 +1,77 @@
+import { ApolloClient } from '@apollo/client/core'
```
znicholasbrown, line 1: This is a general non-Vue apollo client, for use by underlying services and workers.

src/app.js
```diff
@@ -0,0 +1,307 @@
+// Base imports
```
znicholasbrown, line 1: All Vue application-related code has been moved from `main.js` to this file.

src/components/Visualizations/BarChart.vue
```diff
@@ -359,9 +359,17 @@ export default {
 
       context.save()
       context.lineWidth = 2
-      context.strokeStyle = this.computedStyle.getPropertyValue(
-        '--v-appBackground-base'
-      )
+      try {
+        context.strokeStyle = this.computedStyle.getPropertyValue(
+          '--v-appBackground-base'
+        )
```
znicholasbrown, line 365: This was causing some bad background usage.

src/pages/Auth/Logout.vue
```diff
@@ -31,11 +31,10 @@ export default {
     }
   },
   methods: {
-    ...mapActions('auth', ['login']),
     redirectToLogin() {
       this.redirecting = true
       setTimeout(() => {
-        this.login()
+        window.location.href = '/'
```
znicholasbrown, line 37: Since login now happens automatically on application start, we can redirect back to the root directory to start the process.

src/store/tenant/index.js
```diff
@@ -176,30 +157,8 @@ const actions = {
...
-            })
-
-            tokens = tenantToken.data.switch_tenant
-            // Set our new auth token
-            await dispatch('auth/updateAuthorization', tokens, {
-              root: true
-            })
-          }
+          const tokens = await switchTenant(tenant.id)
+          commitTokens(tokens)
```
znicholasbrown, line 161: This is the only except where application logic accesses the auth layer because the application needs to have a way of communicating that it needs to fetch different tokens based on user interaction.

src/vue-apollo.js
```diff
@@ -167,58 +163,6 @@ const authMiddleware = setContext(async (_, { headers }) => {
...
-      if (validRefreshToken) {
-        await store.dispatch('auth/refreshAuthorization')
-      } else if (isAuthenticatedUser) {
-        await store.dispatch('auth/authorize')
-      } else {
-        await store.dispatch('auth/login')
-      }
-    }
-  }
-
```
znicholasbrown, line 221: All of this has been removed in favor of letting the auth layer handle everything related to token refreshing, with the terminal link handling in-flight requests that fail token checks due to expiration.

src/workers/auth.worker.js
```diff
@@ -0,0 +1,165 @@
...
+/*
+    values in the tokens Array<Object> take the form:
+    
+    interface TenantTokens {
+        tenantId: string
+        authorizationToken: string
+        refreshToken: string
+    }
+
+*/
```
znicholasbrown, line 19: This interface isn't quite true yet since we aren't enabling juggling multiple tenant logins at the same time; leaving it here as a scaffold for the future.

vue.config.js
```diff
@@ -99,5 +99,10 @@ module.exports = {
   // Adding this package specically because
   // the source has 2 nullish coalscent references
   // that babel isn't transpiling correctly otherwise
-  transpileDependencies: ['@d3fc/d3fc-axis']
+  transpileDependencies: [
+    '@d3fc/d3fc-axis',
+    'graphql-language-service-interface',
+    'graphql-language-service-parser',
+    'graphql-language-service'
```
znicholasbrown, line 106: These needed to be explicitly transpiled because of [this issue](https://github.com/graphql/graphiql/issues/1852)

zhen0:
I've run this in dev and local for the last few days and the only issue I see is when I try to use the backend switcher from server to cloud - if I'm logged into cloud I briefly see the welcome/name-team screen and then get taken to the dashboard.  If I'm not logged into cloud I get taken to the name-team screen and then it hangs.  I can use the back-button to go back to server a refresh also takes me back to server.

zhen0 (approved):
LGTM @znicholasbrown. We'll address the backend switcher in a follow up PR.

PR #822, opened 2021-05-08
Add recently updated flow tile

zangell44:
PR Checklist:

## Describe this PR

Draft PR for feedback on recently updated flow notification tile. 

A couple things I had trouble with and I'm not sure if I've found the right solution:

1. Hiding the tile correctly (for spacing purposes) if no recent flows are found. Emitting the property from `FlowRecentUpdate` to  `Dashboard` seems to be the best way to do this

2. Updating variables dynamically on queries. Apollo doesn't understand that dates will change between queries, so variables won't be refreshed without some workarounds. The current workaround is to update a property used in query variables during the `update` callback.

Preview:

image

### DIFF
--- a/src/pages/Dashboard/FlowRecentUpdate.vue
+++ b/src/pages/Dashboard/FlowRecentUpdate.vue
@@ -1,0 +1,116 @@
+<script>
+import { mapGetters } from 'vuex'
+
+export default {
+  data() {
+    return {
+      recentlyUpdatedFlow: null,
+      loading: 0,
+      fiveMinutesAgoIsoString: new Date().toISOString()
+    }
+  },
+  created() {
+    this.computeFiveMinutesAgoIsoString()
+  },
+  computed: {
+    ...mapGetters('api', ['isCloud']),
+    ...mapGetters('tenant', ['tenant']),
+    ...mapGetters('user', ['user']),
+    minutesSinceUpdateText() {
+      if (this.recentlyUpdatedFlow === undefined) {
+        return ''
+      }
+      const diffMins = Math.floor(
+        ((new Date() -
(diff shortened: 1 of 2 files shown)


src/pages/Dashboard/FlowRecentUpdate.vue
```diff
@@ -0,0 +1,119 @@
...
+      },
+      variables() {
+        let sortBy = {
+          updated: 'desc'
+        }
+
+        let searchParams = [
+          { archived: { _eq: false } },
+          { updated: { _gte: this.fiveMinutesAgoIsoString } }
+        ]
```
zangell44, line 60: do I need to explicitly add `tenant` to the search params here?

znicholasbrown:
Previews should be working for this 👍

znicholasbrown (changes requested):
Hm I don't think this is quite the correct way to do this.

1) This shouldn't be a dashboard tile; those should be reserved for static content
2) This isn't dismissible by the user, even after clicking "Go to flow"
3) This requires a quite aggressive query to display; I wonder if the global flows query might be used instead? This way we could watch the store and not poll on this component at all

Some suggestions:
Add the `updated` field to the global flows query (`@/graphql/Nav/flows.gql`); watch the store for changes to that
Use [Vuetify snackbars](https://vuetifyjs.com/en/components/snackbars/) for display; by default these are dismissible and survive route changes; we can make use of either local storage or the user settings blob to keep track of which flows (and versions) have been acked by the user 
Use existing date time methods for date formatting; we've got a bunch of useful methods in `@/mixins/formatTime.js` 😄

zhen0:
@zangell44 - can we close this for now?

zhen0:
Discussed with @zangell44 and closing for now as this is no longer a priority

Issue #823, opened 2021-05-10
BUG: Members page doesn't load properly


znicholasbrown:
**Bug Description**

Trying to load the members page is throwing this error:
```
build.umd.js:645 Uncaught TypeError: Cannot read property 'offsetWidth' of undefined
    at Members.vue:126
    at n (build.umd.js:641)
```

**Steps To Reproduce**

1. Visit the members page

**Browsers Tested:**
* [x] Chrome
* [ ] Firefox
* [ ] Safari
* [ ] Edge
* [ ] IE

----------
**Notes**

PR #824, opened 2021-05-10
Members page doesn't load properly

ThatGalNatalie:
PR Checklist:

## Describe this PR
I think this was the issue - It seems like it was trying to reference this.allowedUsers but it wasn't referencing the one from the store but the computed property


znicholasbrown (approved):
Thanks @ThatGalNatalie 

Resolves: #823

PR #825, opened 2021-05-13
Add 3 new tenant-wide actions

znicholasbrown:
PR Checklist:

## Describe this PR
Adds 2 new tenant-wide actions:
- Cancel all runs: calls the `cancel_flow_run` mutation on every run in a `Submitted` or `Running` state
- Pause work queue: stops all new work from being submitted to agents

And updates dashboard visuals for the clear late runs action.

New screens/VUs

**Cancelling in-progress and submitted runs:**
Screen Shot 2021-05-12 at 8 30 28 PM

**Stop all runs InProgressTile overlay:**
Screen Shot 2021-05-12 at 8 30 22 PM

**Clear late runs overlay:**
Screen Shot 2021-05-12 at 8 30 02 PM

**Queue options overlay:**
Screen Shot 2021-05-12 at 8 29 04 PM

**Work queue paused overlay:**
Screen Shot 2021-05-12 at 8 28 45 PM


src/components/SystemActions/CancelAll.vue
```diff
@@ -0,0 +1,170 @@
...
+  computed: {
+    ...mapGetters('tenant', ['tenant']),
+    loading() {
+      return this.loadingKey > 0
+    }
+  },
+  methods: {
+    ...mapActions('alert', ['setAlert']),
+    async cancelAll() {
+      if (!this.flowRuns?.length) return
```
zhen0, line 29: Possibly connected to https://github.com/PrefectHQ/cloud/issues/3429 - this doesn't work but also doesn't show an error in my tenant with 1000+ late runs (which is admittedly rare but we've seen it happen).  Would it be possible to batch by 500 runs?

zhen0: Should we add the same logic to the clear button in the submitted/late runs tile of the flow page?  With this set up one 'clear' button would clear runs while the other (admittedly with a warning dialog) cancels them.

src/App.vue
```diff
@@ -424,6 +426,8 @@ export default {
     <v-main :class="{ 'pt-0': isWelcome }">
       <v-progress-linear absolute :active="loading" indeterminate height="5" />
 
+      <WorkQueueBanner />
```
zhen0, line 429: For me this appears in the same spot at the current alert.  Could we organize them?

zhen0 (changes requested):
This is looking good @znicholasbrown.  A few small nitpicks/comments/nice to haves from me.

src/pages/Dashboard/UpcomingRuns-Tile.vue
```diff
@@ -407,60 +444,52 @@ export default {
...
+      <v-btn
+        v-if="!overlay"
+        small
+        depressed
+        color="primary"
+        text
+        style="z-index: 2;"
+        @click="showOverlay('queue')"
+      >
+        Options
```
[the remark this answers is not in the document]
znicholasbrown: Hm, that wording is intentional for now to avoid the feeling that the "Pause work queue" would happen with just 1 click; in this case it indicates that there's more to do. 

Wrt slowness: I think that's a fairly common use case so let me investigate what's going on there, I'd want the overlay to show regardless of the number of runs.

src/components/SystemActions/CancelAll.vue
znicholasbrown: We can definitely batch though if I'm reading that issue correctly that wouldn't be the cause of this not working; we don't toggle the schedule here to clear these runs, we pass the runs explicitly and loop through them calling the `delete_flow_run` mutation (which is what @cicdw suggests there). However we _do_ still toggle the schedule on the flow page upcoming runs tile, iirc.

I was holding off on making unnecessary changes to the flow page tile to make this easier to review but I'm fine with switching it so they're consistent 👍🏻

src/App.vue
znicholasbrown: I think they should organize themselves by default, are they overlapping?

src/components/SystemActions/CancelAll.vue
znicholasbrown: Once [bulk deletion](https://github.com/PrefectHQ/cloud/pull/3456) is ready we can use that for Cloud, that'll make this far more performant - I'll wait to update this PR though

cicdw: @znicholasbrown I didn't expose the bulk delete route as a GraphQL call -- is that something you're expecting / need?

znicholasbrown: Oh my bad I read your PR wrong - that would certainly improve our ability to perform bulk actions like this (calling `delete_flow_run` in a loop)

znicholasbrown: Flow page tile has been updated for consistency; I can't seem to recreate the issue with the clear button not working so maybe it's best to leave the batching question to a different PR? (this logic uses the same methods as existing, minus schedule toggling)

src/pages/Dashboard/UpcomingRuns-Tile.vue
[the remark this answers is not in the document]
znicholasbrown: Ok found the issue with the slowness, should be much improved now: basically when adding the overlay I used a `v-if`, which completely removes anything under it when the condition isn't met. Using `v-show` prevents us from having to re-render the entire list everytime the overlay is pressed. 👍🏻

src/App.vue
znicholasbrown: Ah nope, didn't realize the material spec dictates that shouldn't be the case. I'll move this one so they shouldn't overlap with our normal ones 👍🏻

znicholasbrown: Screen Shot 2021-05-17 at 7 40 32 PM

Errors and notifications using the existing `Alert` component should now be properly nudged up.

src/components/SystemActions/CancelAll.vue
zhen0: Sounds ok to me.  One to keep an eye on.  (I think we introduced schedule toggling because the cancel all mutation frequently timed out.)

zhen0 (approved):
Thanks for the changes @znicholasbrown - LGTM!

PR #828, opened 2021-05-13
Fix issue setting default tenant in server and bypassing auth

znicholasbrown:
PR Checklist:

## Describe this PR
Fixes an issue where tenant was being set from a default membership, which broke in cases where there was no user object. Also fixes an issue where auth wasn't bypassed correctly.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -12,7 +12,7 @@
 
 ### Bugfixes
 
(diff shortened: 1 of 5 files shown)


zhen0 (approved):
LGTM @znicholasbrown

PR #839, opened 2021-05-17
Run page improvements/fixes

znicholasbrown:
PR Checklist:

## Describe this PR
Fixes issues with pre-populating and resetting job templates and task definitions for run configs, and resets option args when navigating between options, from a stored initial value.


src/components/CustomInputs/MultiLineInput.vue
```diff
@@ -41,17 +43,46 @@ export default {
...
 
         // Use next tick to make sure the json input element exists
         this.$nextTick(() => {
           this.$refs['json-input'].validateJson()
         })
+      } else {
+        if (this.jsonInput === '{}') {
+          this.yamlInput = ''
+        } else {
+          this.yamlInput = stringify(JSON.parse(this.jsonInput))
```
zhen0, line 63: Not a major issue but do we need a try/catch here?  If I add yaml to the json input by mistake and try to switch to yaml, the input vanishes. 
Screen Recording 2021-05-18 at 4.53.10 PM

zhen0 (commented):
One small question/suggestion @znicholasbrown

src/components/CustomInputs/MultiLineInput.vue
zhen0: https://user-images.githubusercontent.com/40272060/118722313-03e55400-b7fa-11eb-876a-ebd9dd782167.mov

znicholasbrown: Hm yeah I think a try catch makes sense here - will add that

znicholasbrown: Updated 👍🏻

zhen0 (approved):
Thanks for the changes @znicholasbrown

PR #841, opened 2021-05-17
allowed users error

ThatGalNatalie:
PR Checklist:

## Describe this PR
Set the initial value of this.users and this.invitations was undefined on mount 
Made this.allowedUsers a computed prop to use the nullish coalescing operator /check for total allowed users

### DIFF
--- a/src/pages/TeamSettings/Members.vue
+++ b/src/pages/TeamSettings/Members.vue
@@ -395,7 +395,7 @@ export default {
       :error="inviteError"
       :loading="isInvitingUser"
       :disabled="
-        totalUsers >= allowedUsers || !inviteFormValid || isInvitingUser
+        totalUsers >= totalAllowedUsers || !inviteFormValid || isInvitingUser
       "
       :dialog-props="{
         'max-width': '600'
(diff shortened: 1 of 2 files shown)


znicholasbrown:
This removed the warning but I think in the wrong way - testing on a tenant with 10 allowed users (from the license terms) but more than 10 users already in the tenant and no warning is displayed

znicholasbrown (dismissed):
This removed the warning but I think in the wrong way - testing on a tenant with 10 allowed users (from the license terms) but more than 10 users already in the tenant and no warning is displayed

src/pages/TeamSettings/Members.vue
```diff
@@ -81,8 +81,11 @@ export default {
     ...mapGetters('tenant', ['tenant']),
     ...mapGetters('user', ['user']),
     ...mapGetters('license', ['license', 'hasPermission', 'allowedUsers']),
+    totalAllowedUsers() {
+      return this.allowedUsers() ?? Infinity
+    },
     insufficientUsers() {
-      return this.users >= this.allowedUsers() ?? Infinity
+      return this.users >= this.totalAllowedUsers
```
ThatGalNatalie, line 88: Looks like `this.users` is undefined. Will fix now

znicholasbrown (approved):
Works great, thanks for the changes @ThatGalNatalie 👍🏻

PR #846, opened 2021-05-18
Improve usability of the quick run button

znicholasbrown:
PR Checklist:

## Describe this PR
This PR removes the automatic routing of the quick run button and introduces a new toast system for stackable global toasts with a straightforward CRUD system. It should now be a lot easier to create a bunch of runs with default parameters!

https://user-images.githubusercontent.com/27291717/118725558-19f51380-b7fe-11eb-8692-6a220b90c261.mov

### DIFF
--- a/src/pages/Flow/Actions.vue
+++ b/src/pages/Flow/Actions.vue
@@ -85,7 +85,11 @@ export default {
     }
   },
   methods: {
-    ...mapActions('alert', ['setAlert']),
+    ...mapActions('alert', [
+      'setAlert',
+      'addNotification',
+      'updateNotification'
+    ]),
     _changeVersion(val) {
       if (val) {
         let query = { ...this.$route.query, version: val }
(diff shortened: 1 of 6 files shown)


madkinsz:
Sweet. I hit this exact pain point just a couple days ago.

ThatGalNatalie (approved):
lgtm!

src/pages/Flow/Actions.vue
```diff
@@ -117,45 +128,58 @@ export default {
...
-        this.quickRunErrorAlert()
+        await this.updateNotification({
+          id: notificationId,
+          notification: {
+            color: 'error',
+            loading: false,
+            text: 'There was a problem creating your run',
+            subtext: err?.message?.includes('archived')
+              ? 'This flow version is archived - you likely have a newer version of your flow'
+              : 'Error',
```
zhen0, line 174: Maybe this is one for wider discussion but I prefer adding the error message passed back by the API.  They're now written in a way that is meant to be user friendly and help us and users with de-bugging.  Either way, I'm not sure 'Error' is helpful - if we don't include the actual error message maybe just leave blank?

zhen0 (commented):
This looks really good @znicholasbrown - one comment about error messages.

src/pages/Flow/Actions.vue
znicholasbrown: Yeah I agree in most cases wrt API error messages but less so in application errors that we don't explicitly throw. In this particular instance though, the API is returning almost the same message but with an id embedded, which isn't necessary to show the user (they're already on the flow page!)

zhen0 (approved):
LGTM @znicholasbrown

Issue #109, opened 2020-06-08
Placing Prefect UI behind proxy


oleksandr:
## Current behavior

Currently it's pretty much impossible to put Prefect UI container behind proxy and make it serve under "/some/other/path" but not root "/".

## Proposed behavior

It would be great to implement either of options:
 1. Pass env var (e.g. `PROXY_ADDRESS_FORWARDING`) that will respect all incoming `X-Forwarding-*` headers and adjust behaviour.
 2. Pass env var (e.g. `BASE_URL`) to communicate to the static app it's not in the root path.

majikman111:
Is this features currently being implemented? I am using Prefect behind a reverse proxy and previously was able to create a custom docker image as described in PR #450. But as of version 2021-01-05 on-site builds are not possible due to embedded private keys. Now I cannot update the UI as I cannot serve it from behind a reverse proxy.

dwissm1:
+1 desperately needed.

igonro:
Hello @oleksandr @majikman111 @dwissm1 

Were you able to make it work behind a proxy? I am having some issues to make it work. Can you help me in [this discussion](https://github.com/PrefectHQ/prefect/discussions/5376)?

Thanks!

TonyTromp:
+1

igonro:
I know this isn't a suitable solution for all the scenarios, but someone may find this useful.

If you are coming to this issue because you have a Prefect on a server without GUI and you want to use Prefect UI, the easiest solution is to connect to the server using SSH port forwarding (remember to enable it for Prefect UI and Prefect Server), this way you can use Prefect UI from external PC.

Example:
1. Write something like this in the ssh config file.
```yaml
# ~/.ssh/config
Host prefect-port-forwarding
    HostName YOUR_SERVER_IP
    User YOUR_USER
    LocalForward 8080 localhost:8080
    LocalForward 4200 localhost:4200
```
2. Run `ssh prefect-port-forwarding`.
3. Open a tab in your browser and go to `http://localhost:8080`.
4. Set the `Prefect Server GraphQL endpoint` to `http://localhost:4200/graphql`.

And now you can use Prefect UI in your PC even if Prefect is in a server. 😄 

It's a pretty simple and maybe obvious solution, but we were going crazy trying to place Prefect UI behind Nginx, and this simple solution satisfied our needs.

PR #849, opened 2021-05-19
Allow configurable base url and relative public assets path

znicholasbrown:
PR Checklist:

## Describe this PR
Resolves: #109 
Resolves: #450 

~~Right now this will only work with direct modification of `settings.json` - will need a corresponding Core PR to pass this setting to the build.~~
Core PR: https://github.com/PrefectHQ/prefect/pull/4557


ThatGalNatalie (approved):
lgtm!

zhen0 (commented):
One question about v3 vs v4 vue router @znicholasbrown

zhen0 (approved):
Version issue resolved.  We just happen to be using the same name as the new method in V4.

PR #852, opened 2021-05-19
Add agents

zhen0:
PR Checklist:

## Describe this PR
Adds agent screens with flow run history, submittable runs, runs in progress and agent config info.


znicholasbrown:
Some initial comments:
- Cleaning up agents doesn't seem to be working as intended; the first time I press the button nothing seems to have happened and the second time only greyed-out agents were cleaned up. It doesn't seem like unhealthy agents are all removed, maybe because they have submittable or late runs?
- This dialog box is really big:
Screen Shot 2021-05-27 at 1 47 53 PM
- When filters are applied, the toast message blocks content:
Screen Shot 2021-05-27 at 1 50 58 PM
- Is the "Recent Runs" bullet point pointing to the chart below? If so, that isn't super clear since it's in the same format as the single-line information above (submittable runs and last queried)
Screen Shot 2021-05-27 at 1 51 37 PM
- Query times don't seem to be counting up and the tooltip timestamp is incorrect (I think; this screenshot was taken at 13:41 EST):
Screen Shot 2021-05-27 at 1 41 16 PM

On the agent page:
- This is overall looking nice! The numbers in this screenshot correspond to the numbered feedback:
Screen Shot 2021-05-27 at 1 54 21 PM
1. We should use the agent name or a more human-friendly default for unnamed agents instead of the id
2. This is a really large amount of padding - it should probably match the padding on the flow run page
3. The typography here makes it difficult to parse key:value; a design similar to the overview tile on flow run pages would help a lot (floated key:values with differing font weights)
4. (this applies to the agent overview cards on `/agents`): when there are no labels, we should display something to that effect, instead of having nothing at all. Maybe "No labels" with `text--disabled` applied?
5. Duplicate wording on this tile; since it's on the agent page, it can be inferred that the config is an agent config. I'm also thinking this tile feels a little unimportant. Maybe using a similar top-right tab style to the flow page details would be better? Then we can trim the wording a lot, with just a `Config` tab and `No config` when there isn't one.
Screen Shot 2021-05-27 at 2 04 47 PM


Small note: the styling for upcoming runs (on which the submittable runs tile is based) has changed a bit for clarity, so this design should be updated as well.

zhen0:
@znicholasbrown - This one is waiting for review.

znicholasbrown (changes requested):
A few pieces of feedback that weren't addressed from the last review:
- [x] the padding between tiles is uneven
- [ ] The typography on the overview tile makes it difficult to parse key:value; a design similar to the overview tile on flow run pages would help a lot (floated key:values with differing font weights)
- [x] when no config is present we can say "No config" instead of "no agent config is present on this agent", which feels redundant


And some new feedback:
- [x] the first row agent page tiles are too tall; on a medium dpi laptop they have a lot of unused whitespace and hide the next row below the fold:
Screen Shot 2021-06-07 at 9 31 58 PM


- [x] I think this would feel better if you used `/agent` as the root page and then added breadcrumbs on the individual agent pages to get back to the primary agents screen (so `/agent/:id?` for the route, showing all the agents tiles when no id is provided)
Screen Shot 2021-06-07 at 9 32 46 PM

- [x] the agent timeline is missing flow names
Screen Shot 2021-06-07 at 9 36 00 PM

- [x] the space between recent runs and last queried could use a visual break - maybe a horizontal rule would help here?
Screen Shot 2021-06-07 at 9 40 50 PM

- [x] agents with no recent runs just show a line; it'd be good to have some text letting the user know there are no runs like on the flow/dashboard pages' timelines
Screen Shot 2021-06-07 at 9 48 29 PM

- [x] instead of 720 minutes, let's put 12 hours here
Screen Shot 2021-06-07 at 9 56 52 PM

znicholasbrown:
Another note: I think this can be improved a bit; we probably don't need to hide the subpage nav bar when there are no agents since this looks like a UI error and not something actionable for the user. I'd also change the wording a little:
> You have no agents querying for runs.
> For help setting up an agent, check out the agent [docs] or [tutorial].

Screen Shot 2021-06-09 at 10 28 22 PM

zhen0:
Made those changes @znicholasbrown - ready for review.

src/App.vue
```diff
@@ -134,6 +135,11 @@ export default {
     isDark(val) {
       this.$vuetify.theme.dark = val
     },
+    refetchAgents(val) {
+      if (val) {
+        this.$apollo.queries.agents.refresh()
+      }
+    },
```
znicholasbrown, line 142: I'm a little confused what this watcher is doing that `this.$globalApolloQueries['agents'].refetch()` wouldn't do without a new store pattern

src/graphql/Agent/FlowRuns.gql
```diff
@@ -0,0 +1,17 @@
...
+    labels
+    scheduled_start_time
+    version
+    flow {
+      id
+      name
+      schedule
+      environment
+      is_schedule_active
+  }
```
znicholasbrown, line 15: Since we have all flows in the data store, this query would be a lot more performant by grabbing the `flow_id` and tying the data together instead

src/pages/Agents/AgentRunsInProgress.vue
```diff
@@ -0,0 +1,121 @@
+<script>
```
znicholasbrown, line 1: It looks like this component isn't used, can it be removed?

src/router.js
```diff
@@ -223,6 +223,23 @@ export const routes = [
...
+  // ---------------------------//
+  //
+  // Agents
+  //
+  // ----------------------------//
+  {
+    name: 'agents',
+    path: '/:tenant?/agent',
+    component: () =>
+      import(/*webpackChunkName: "calendar" */ '@/pages/Agents/Agents')
```
znicholasbrown, line 235:
```suggestion
      import(/*webpackChunkName: "agents" */ '@/pages/Agents/Agents')
```

src/router.js
```diff
@@ -223,6 +223,23 @@ export const routes = [
...
+  },
+  {
+    name: 'agent',
+    path: '/:tenant?/agent/:id',
+    component: () =>
+      import(/* webpackChunkName: "flow" */ '@/pages/Agents/AgentPage.vue')
```
znicholasbrown, line 241:
```suggestion
      import(/* webpackChunkName: "agent" */ '@/pages/Agents/AgentPage.vue')
```

znicholasbrown (changes requested):
Some small comments but mostly looks good 😄

znicholasbrown (approved):
Thanks for the changes @zhen0 !

Issue #860, opened 2021-05-26
BUG: When I accept an invitation through the team switcher I need to refresh the page to see the new team


dylanbhughes:
**Bug Description**

**Steps To Reproduce**

1. Invite a user to a team
2. Navigate to the team switcher
3. Accept the invitation
4. New team doesn't appear in the list

**Browsers Tested:**
* [X] Chrome
* [ ] Firefox
* [ ] Safari
* [ ] Edge
* [ ] IE

----------
**Notes**

PR #862, opened 2021-05-26
Fix routing from invitation

znicholasbrown:
PR Checklist:

## Describe this PR
This PR addresses the following:
- Membership and invitation tables not updating
- Teams not updating after accepting an invitation (Resolves: #860)
- Access denied when the IDP client application throws
- `null` tenant id on the login token, defaults to the first tenant now and throws access denied if no tenants exist

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -8,7 +8,7 @@
 
 ### Bugfixes
 
-- None
+- Various bugfixes related to null tenants and team invitaitons - [#862](https://github.com/PrefectHQ/ui/pull/862)
 
 ## 2021-05-25
 
--- a/src/App.vue
+++ b/src/App.vue
@@ -226,6 +226,23 @@ export default {
       }
     }
 
+    this.$globalApolloQueries['tenants'] = this.$apollo.addSmartQuery(
+      'tenants',
+      {
+        query: require('@/graphql/Tenant/tenants.js').default(this.isCloud),
+        skip() {
+          return (this.isCloud && !this.isAuthorized) || !this.connected
+        },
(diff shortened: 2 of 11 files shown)


zhen0 (changes requested):
This all looked good to me until I tried it out.  I removed my email from an old tenant and then invited that user to join.  The invitation got sent ok and I got taken to the login and then the correct page with the invitation.  However, when I clicked join I got the dreaded "cannot read property 'data' of undefined'.  Clicking on 'Back to the Dashboard' worked but I still hadn't joined my new tenant.  Clicking on the tenant switcher left me in spinning/loading limbo.  I dug around a little and the issue seems to come from the accept-membership-invitation mutation which is not returning anything.  That mutation worked fine in the IAPI.  Happy to screen-share give more details if needed.  May not be directly related to your PR but I think it's connected to the issue you're fixing.

zhen0:
We discussed and tried to replicate but can't. Happy to approve this PR as doubt it's directly related and keep investigating separately.

znicholasbrown:
👍🏻  Fixed some issues routing to the invitation acceptance page on application start and some small visual stuff on that page so should be good to go now.

zhen0:
I'm seeing a flicker when i delete a user on this branch.  They are removed, come back and then are finally fully removed.

znicholasbrown:
Ok that should be fixed now @zhen0  - looks like the `no-cache` policy wasn't being respected properly, so switched it to `network-only`.

zhen0:
Thanks for the switch @znicholasbrown LGTM!

zhen0 (approved):

PR #863, opened 2021-05-27
Templating Task Name

### LINKED ISSUE #853: BUG: Templating Task Names Updates Task List in UI but not Schematic
**Bug Description**

When using templates to name tasks, the UI shows different names for the schematic and the task list in the Overview. The schematic shows the untemplated name while the overview shows the correctly templated name.

**Steps To Reproduce**

Here is a code snippet:
```python
from prefect.storage import GCS

@task(task_run_name="{val}")
def compute(val):
    pass
with Flow("template-example", storage=GCS(bucket="prefect-flows-poc2021")) as flow:
    compute(val="hello")

flow.register(project_name="fraud_model", labels=["gcp"])
```

Schematic will say "compute" while the task list in the overview will say "hello"

imageimage

ThatGalNatalie:
PR Checklist:

## Describe this PR
Should resolve #853

### DIFF
--- a/src/components/Schematics/Schematic-Flow.vue
+++ b/src/components/Schematics/Schematic-Flow.vue
@@ -376,7 +376,7 @@ export default {
         let task = taskMap[node.id]
         // This ensures the id isn't incorrectly overwritten
         let parsedTask = task.task ? { ...task, ...task.task } : task
-        parsedTask = { ...parsedTask, ...{ name: task?.name } }
+        parsedTask.name = task?.name ? task.name : parsedTask.name
 
         node.data = parsedTask
       })
--- a/src/components/Schematics/Schematic-Flow.vue
+++ b/src/components/Schematics/Schematic-Flow.vue
@@ -376,6 +376,7 @@ export default {
         let task = taskMap[node.id]
         // This ensures the id isn't incorrectly overwritten
         let parsedTask = task.task ? { ...task, ...task.task } : task
+        parsedTask = { ...parsedTask, ...{ name: task?.name } }
 
         node.data = parsedTask
       })
--- a/src/components/Schematics/Schematic-Node.vue
+++ b/src/components/Schematics/Schematic-Node.vue
@@ -209,7 +209,10 @@ export default {
         </v-avatar>
         {{ nodeData.data.task.name }}
       </div>
-      <div class="text-h6 text-truncate font-weight-bold" :style="titleStyle">
+      <div
+        class="text-h6 text-truncate font-weight-bold mt-4"
+        :style="titleStyle"
+      >
         <v-avatar
           v-if="isResource || isParameter"
           color="accentOrange"
--- a/src/components/Schematics/Schematic-Node.vue
+++ b/src/components/Schematics/Schematic-Node.vue
@@ -191,7 +191,7 @@ export default {
   >
     <div class="ml-12 node-content" style="width: calc(100% - 6rem);">
       <div
-        v-if="nodeData.data.task"
+        v-if="nodeData.data.task && !nodeData.data.mapped"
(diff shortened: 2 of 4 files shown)


znicholasbrown:
I think we can show both the task run name and the task name, something like this for flow runs:
```
<<task name>> (test-subtitle-2 font-weight-light)
<<task run name>> (text-h5)
<<duration>> (text-subtitle-2)
```
but then for mapped tasks we could do:
```
<<task name>> (text-h5)
<<number of finished runs / total number>> (text-subtitle-2 font-weight-light)
<<duration>> (text-subtitle-2)
```

(note that those are relative font classes, I didn't check what the current ones are or whether they're even class controlled)

We'd want to keep the static graph as-is.

src/components/Schematics/Schematic-Flow.vue
```diff
@@ -376,6 +376,7 @@ export default {
         let task = taskMap[node.id]
         // This ensures the id isn't incorrectly overwritten
         let parsedTask = task.task ? { ...task, ...task.task } : task
+        parsedTask = { ...parsedTask, ...{ name: task?.name } }
```
znicholasbrown, line 379: You shouldn't need to use 2 more spreads here, this will work:
```
parsedTask.name = task?.name ? task.name : parsedTask.name
```

znicholasbrown:
It looks like we're still showing special class type icons on the task name line. This view might be confusing with duplicates when runs aren't explicitly named, but perhaps not. Maybe making bumping the overline text down a little would help?
Screen Shot 2021-06-01 at 12 11 41 PM

znicholasbrown:
I think there might be some unpushed changes, this looks largely the same as before:
Screen Shot 2021-06-04 at 2 42 08 PM
Note that the special class type icons are still duplicated on the task name line.

On the finished runs portion, that can be said more succinctly like this:
```
current/total runs complete
```

and when zoomed out far enough that the chart/duration are no longer showed, probably just `current/total`.

I think this has introduced a small bug with node scaling/placement (note that terminal nodes have shifted down somewhat and are no longer centered on the end of the edge):
this branch:
Screen Shot 2021-06-04 at 2 44 30 PM
master:
Screen Shot 2021-06-04 at 2 44 34 PM

src/components/Schematics/Schematic-Node.vue
```diff
@@ -204,6 +223,21 @@ export default {
...
+        <div v-if="showDetails">
+          {{
+            mappedChildren.state_counts['Success']
+              ? mappedChildren.state_counts['Success']
+              : 0
+          }}/{{ totalRuns }} {{ totalRuns > 1 ? 'runs' : 'run' }} complete
+        </div>
+
+        <div v-else class="mt-2">
+          {{ mappedChildren.state_counts['Success'] }}/{{ totalRuns }}
```
znicholasbrown, line 237: Don't we want to count all the runs that have entered any finished state instead of just `Success`?

src/components/Schematics/Schematic-Node.vue
```diff
@@ -116,6 +117,17 @@ export default {
...
+    totalRuns() {
+      if (!this.mappedChildren) return
+
+      return Object.values(this.mappedChildren.state_counts).reduce(function(
+        a,
+        b
+      ) {
+        return a + b
+      },
+      0)
```
znicholasbrown, line 130: I'm not sure this is necessary; don't we have this number in the `serialized_state.n_map_states` field on the task run?

znicholasbrown (changes requested):
This looks good, a few small code comments

znicholasbrown (approved):
Nice, thanks for the updates @ThatGalNatalie 👍🏻

PR #881, opened 2021-06-02
Roles

zhen0:
PR Checklist:

## Describe this PR
- Adds a custom roles page


zhen0:
Not at all neat notes from discussion with Nicholas: 

- [ ] Make the delete only show on hover, 

- [ ] user friendly permission types- group, makes that fit what they do, format, 

- [ ] smaller button, show new role as soon as hot button- in-line editable, 
- [ ] custom roles list section scrollable, 
- [ ] more prominent highlight for role name selection, 
- [ ] edit button to enable editing- actions at top, 
- [ ] hide disabled options/permissions; 
- [ ] typography- match management - lighter font, 
- [ ] delete role button like notifications and membership and tenant switcher- use that component, 
- [ ] loading, 
- [ ] stop whole page scroll, 
- [ ] plus button next to header for new role; 
- [ ] don’t capitalize every word for default roles (or actually use same formatting and names as on members page?)

zhen0:
Next set of feedback comments: 

- [ ]  Make default roles disabled instead of read only and add a header that notes they can not be modified
- [ ] If we close edit, go back to the default role
- [ ] remove the user permissions list
- [ ] Font weight light for headers (similar styling to the account page)
- [ ] Adjust the margin and padding on the checkbox header rows and try centering the text
- [ ] Move the role name edit slot to the top of the custom roles list
- [ ] Add elevation/border to the reset/update action bar 
- [ ] Make rest and update buttons smaller
- [ ] Add an edit button and disable all check boxes until edit is clicked then switch edit button to reset/update
- [ ] Make 'reset' 'cancel'
- [ ] Remove the 'add' icon when in process of adding a role

znicholasbrown (approved):
LGTM @zhen0!

PR #887, opened 2021-06-07
KV Copy

ThatGalNatalie:
PR Checklist:

## Describe this PR
Currently, clicking the "copy" button in the key value adds quotation marks to the value if it's not an object. This should only stringify the value if it's an object.


znicholasbrown (approved):
LGTM 👍🏻

PR #908, opened 2021-06-22
members and service account permissions and formatting 

ThatGalNatalie:
PR Checklist:

## Describe this PR
This checks for permissions on the members and service account pages as well as formats the roles


src/pages/TeamSettings/Members/Invitations-Table.vue
```diff
@@ -196,13 +196,21 @@ export default {
...
+              : roleColorMap[item.role] || 'secondaryLight'
+          "
+        >
+          {{ roleMap[item.role] || item.role }}
         </v-chip>
       </template>
 
       <!-- ACTIONS -->
-      <template v-if="isTenantAdmin" #item.action="{ item }">
+      <template v-if="permissionsCheck" #item.action="{ item }">
```
zhen0, line 213: I believe this should be hasPermission('delete', 'membership-invitation')

src/pages/TeamSettings/Members/Members-Table.vue
```diff
@@ -294,7 +313,13 @@ export default {
           </template>
           Modify this user's role
         </v-tooltip>
-        <v-tooltip bottom>
+        <v-tooltip
+          v-if="
+            hasPermission('feature', 'basic-rbac') &&
```
zhen0, line 318: Tenants without rbac may still need to delete members

src/pages/TeamSettings/Members.vue
```diff
@@ -222,15 +239,15 @@ export default {
...
+      <span v-if="permissionsCheck">
         View your team's members, manage permissions, and send invitations
       </span>
       <span v-else data-cy="non-admin-message">
         View the other members of {{ tenant.name }}
       </span>
     </template>
 
-    <template v-if="isTenantAdmin" #cta>
+    <template v-if="permissionsCheck" #cta>
```
zhen0, line 250: This should be hasPermission('create', 'membership-invitation')

src/pages/TeamSettings/Members/Members-Table.vue
```diff
@@ -268,14 +273,28 @@ export default {
...
         </v-chip>
       </template>
 
       <!-- ACTIONS -->
-      <template v-if="isTenantAdmin" #item.actions="{ item }">
-        <v-tooltip v-if="hasPermission('feature', 'basic-rbac')" bottom>
+      <template v-if="permissionsCheck" #item.actions="{ item }">
+        <v-tooltip
+          v-if="
+            hasPermission('feature', 'basic-rbac') &&
```
zhen0, line 293: May need to check for feature:custom-role also.  (Or remove the basic-rbac check)

zhen0 (changes requested):
My main comment is (as we discussed) permission checks can be more granular. Check in local and the error handling and refetching look much better.

src/pages/TeamSettings/Members/Members-Table.vue
znicholasbrown: i think `basic-rbac` is fine for now - this needs to have the basic-rbac check for sure and since this isn't creating a custom role there shouldn't be any blocker in the backend.

znicholasbrown (changes requested):
A few small comments in addition to @zhen0's. I think we need to introduce a bypass in Server for all permissions but I can introduce that in another PR

src/pages/TeamSettings/Members.vue
```diff
@@ -373,7 +391,7 @@ export default {
       <!-- PENDING INVITATIONS TABLE -->
       <v-tab-item value="pending" eager>
         <InvitationsTable
-          :is-tenant-admin="isTenantAdmin"
+          :permissions-check="permissionsCheck"
```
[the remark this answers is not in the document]
ThatGalNatalie: Since the invitations table currently only contains the revoke invitation button, I was thinking of hasPermission('delete', 'membership-invitation')?

src/pages/TeamSettings/Members/Members-Table.vue
zhen0: Ah Ok. Does that mean that all licenses that have custom-role will also have basic-rbac?

src/pages/TeamSettings/Members.vue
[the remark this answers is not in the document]
znicholasbrown: I think that makes sense; ideally this would be `read:membership-invitation` but I think that doesn't exist

src/pages/TeamSettings/Members/Members-Table.vue
znicholasbrown: That's something I'll raise cause they should

znicholasbrown (approved):
Thanks for the changes @ThatGalNatalie - LGTM

PR #916, opened 2021-06-23
schedule toggle 

ThatGalNatalie:
PR Checklist:

## Describe this PR
Currently, the `disableToggle` check is returning true if a user has create or delete run permissions. This changes it to return true only if they don't have create or delete permissions.


znicholasbrown (approved):
👍🏻 LGTM

PR #918, opened 2021-06-24
Use role id instead of name in members page invite mutation

znicholasbrown:
PR Checklist:

## Describe this PR
Fix an issue where we were passing a role name instead of id in the `role_id` slot of the `create_membership_invitation` mutation.


zhen0 (approved):
Thanks for the update @znicholasbrown!

Issue #885, opened 2021-06-04
BUG: Schedule Config page show wrong time or Timezone


raphael-riel:
**Bug Description**
![image](https://user-images.githubusercontent.com/474697/120845062-474ffa00-c53e-11eb-8cbc-4b3a130b0b6a.png)

![image](https://user-images.githubusercontent.com/474697/120845069-47e89080-c53e-11eb-8c0a-b55dd76c6e64.png)

Should be either 12 UTC, or 8 EDT (As in the upcoming run panel)

**Steps To Reproduce**

See above.

**Browsers Tested:**
* [ ] Chrome
* [x] Firefox
* [ ] Safari
* [ ] Edge
* [ ] IE

Issue #813, opened 2021-05-06
BUG: Default Parameters label on schedules is confusing


zhen0:
**Bug Description**

The default parameters label on schedules in the UI shows up when I create a schedule and do not alter the params and when I create a schedule and do alter the params.  This leaves me uncertain what it's for.  More here: https://prefect-community.slack.com/archives/C0192RWGJQH/p1620102367208800

**Steps To Reproduce**

1. Create a schedule on a flow with parameters. Update a param.  The blue 'default parameter' label shows. 
2. Create a schedule don't add any parameters.  The blue label still shows. 

**Browsers Tested:**
* [ ] Chrome
* [ ] Firefox
* [ ] Safari
* [ ] Edge
* [ ] IE

----------
**Notes**

I wonder if the label should say 'schedule parameters' and only show when schedule parameters are added.

PR #922, opened 2021-06-24
Schedule timezone

ThatGalNatalie:
PR Checklist:

## Describe this PR
Should include a fix for #929,  #928, #885, and #813

### DIFF
--- a/src/pages/Flow/Settings/Schedules.vue
+++ b/src/pages/Flow/Settings/Schedules.vue
@@ -336,6 +336,10 @@ export default {
                 <div v-show="selectedTab === 0">
                   <ClockForm
                     :param="parameter"
+                    :timezone="
+                      this.timezone ||
+                        Intl.DateTimeFormat().resolvedOptions().timeZone
+                    "
                     @cancel="selectedClock = null"
                     @confirm="createClock"
                   />
--- a/src/pages/Flow/Settings/Schedules.vue
+++ b/src/pages/Flow/Settings/Schedules.vue
@@ -239,16 +239,13 @@ export default {
       }
     },
     timezoneVal(clock) {
-      let tz =
-        clock.scheduleType == 'flow'
-          ? 'UTC'
-          : this.timezone || Intl.DateTimeFormat().resolvedOptions().timeZone
-      if (clock?.timezone) {
-        tz = clock?.timezone
-      } else if (clock?.start_date?.tz) {
-        tz = clock?.start_date?.tz
+      if (clock.scheduleType == 'flow') {
+        return clock?.start_date?.tz || 'UTC'
+      } else if (this.clocks[0] && clock.scheduleType == 'flow-group') {
+        return this.clocks[0]?.timezone || this.clocks[0]?.start_date?.tz
       }
-      return tz
+
+      return this.timezone || Intl.DateTimeFormat().resolvedOptions().timeZone
     },
     paramVal(clock) {
       if (!clock) {
@@ -335,6 +332,7 @@ export default {
 
                 <div v-show="selectedTab === 0">
                   <ClockForm
+                    :flow-group-clocks="flowGroupClocks"
                     :param="parameter"
                     :timezone="
                       this.timezone ||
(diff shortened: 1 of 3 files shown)


src/pages/Flow/Settings/Schedules.vue
```diff
@@ -238,7 +239,10 @@ export default {
       }
     },
     timezoneVal(clock) {
-      let tz = this.timezone || Intl.DateTimeFormat().resolvedOptions().timeZone
+      let tz =
+        clock.scheduleType == 'flow'
+          ? 'UTC'
+          : this.timezone || Intl.DateTimeFormat().resolvedOptions().timeZone
```
znicholasbrown, line 245: This seems like it'll change the timezone instead of updating the display for the user selected timezone, am I understanding correctly?

ThatGalNatalie: This method is used in two places, one is the `CronClock` component which is used for abbreviating and the other is the `ClockForm` component for editing. For editing, this method should only preselect the timezone so I don’t think it would change the timezone but maybe I’m missing something? 

I added the `clock.scheduleType == 'flow'` to differentiate between what default should be used as it was defaulting to the user’s settings timezone before

zhen0: Could we make that pre-selected timezone a bit more obvious by showing it in the v-select- such as happens in the user profile?  

What I see in the user profile:

Screen Shot 2021-06-25 at 3 15 17 PM

What I see in the schedules page (note it does not tell me what the timezone is):

Screen Shot 2021-06-25 at 3 17 46 PM

zhen0 (dismissed):
This addresses the basic ticket issue but because there's a difference between the UI and backend default timezone I think we need to make the default timezone more clear in the schedule settings page selector.

ThatGalNatalie:
> This addresses the basic ticket issue but because there's a difference between the UI and backend default timezone I think we need to make the default timezone more clear in the schedule settings page selector.

Agree. I think it'll be more clear if we show the pre-selected timezone on creation too. As it defaults to their timezone in the user profile or local timezone if they don't select one anyways.

zhen0:
Thanks for the changes @ThatGalNatalie - this looks good to me.

src/pages/Flow/Settings/Schedules.vue
```diff
@@ -517,6 +530,29 @@ export default {
                       </template>
                       This schedule overrides default parameters.
                     </v-tooltip>
+
```
zhen0, line 533: I like this warning but could we make it clearer at a glance that this schedule will not be used - maybe reduce the opacity of flow schedule cards to show they are disabled?

zhen0 (changes requested):
Some comments in the code and also a request that we include the timezone selector in the main/non-advanced section of the form - that also creates cronclocks and so has a timezone attached and it would be nicer for users to be able to see which timezone is being used as it's not immediately clear that it would be the timezone of an existing clock as opposed to UTC or the UI default.

src/pages/Flow/Settings/Schedules.vue
```diff
@@ -557,6 +600,11 @@ export default {
                           >
                             This schedule was set in your Flow's code so it
                             can't be modifed.
+
+                            <div v-if="hasFlowGroupSchedule" class="mt-2">
```
zhen0, line 604: This alert still looks very big - can we try a different font - maybe caption?

src/pages/Flow/Settings/Schedules.vue
```diff
@@ -387,6 +401,7 @@ export default {
         :class="{ 'grid-container-large': selectedClock === i }"
       >
         <v-card
+          :disabled="hasFlowGroupSchedule && clock.scheduleType == 'flow'"
```
zhen0, line 404: Disabling stops the tooltips from working so perhaps isn't the best option here.  I think reducing the opacity would work better.

zhen0 (dismissed):
One very small change - the rest looks good to me.

zhen0 (approved):
Thanks @ThatGalNatalie - LGTM!

PR #930, opened 2021-07-01
Audit logs

znicholasbrown:
PR Checklist:

## Describe this PR
Adds a new screen for audit logs as well as new components for a larger logs refactor


zhen0 (commented):
A few initial thoughts & comments:
- I don't see any permission checks and my Read only user cloud access the page but got a message that there were no logs.  Should there be permissions?
- If I switch tenants, my logs don't update. 
- Can we show some sort of indicator when a filter is applied.  This is really only relevant for the log level filter as other filter details are also in the header bar so an alternative would be to show a log level chip?
- Should I be able to see Flow logs?  Not sure if there's none available or I don't have the right permissions (see permissions comment above and note about text color below.)
- In the choice of log type (user, flow etc) the text is very light grey and I'm not sure if it's disabled or just light grey.  I think if we fix permissions that should be less of an issue but do we need the text to be so light here?
Screen Shot 2021-07-01 at 10 54 45 AM
- Is it a deliberate choice not to use standard page headers like the dashboard, agents, & account have? 
- The download and filter buttons are functional but seem a bit grey and have no text formatting - could we try something like the Github Security Log export button - but maybe a more prefect color?
Screen Shot 2021-07-01 at 11 12 10 AM

ThatGalNatalie:
Some initial thoughts:

- If I hover over the filters below, I don’t get the pointer for the cursor. Could we add that?  
curosr

- If I search and no logs are found, is there a way to bring me back to view all my logs without having to refresh the page? 
- If I change either the start or end date in the filter section to a  different date and then click reset, it doesn’t reset for me
- Could we format this time here or does it need to be in timestamp form?
time
- Could we highlight the the flow name in the logs?
flow-name
- This might just be me but the numbers are kinda hard to see when dark mode is enabled
time-in-dark-mode

znicholasbrown:
Closing this for now

PR #935, opened 2021-07-07
Enable the quick run button for flows with required parameters

znicholasbrown:
PR Checklist:

## Describe this PR
Enables the quick run button for flows with required parameters with defaults on the flow object and not the flow group object

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -11,6 +11,7 @@
 - Should include a fix for [#929](https://github.com/PrefectHQ/ui/issues/929), [#928](https://github.com/PrefectHQ/ui/issues/928), [#885](https://github.com/PrefectHQ/ui/issues/885), and [#813](https://github.com/PrefectHQ/ui/issues/813) - [#922](https://github.com/PrefectHQ/ui/pull/922)
 - Fix role highlight in dark mode - [#925](https://github.com/PrefectHQ/ui/pull/925)
 - Fix indentation error in doc code snippet - [#932](https://github.com/PrefectHQ/ui/pull/932)
+- Enables the quick run button for flows with required parameters with defaults on the flow object and not the flow group object - [#935](https://github.com/PrefectHQ/ui/pull/935)
 
 ## 2021-06-24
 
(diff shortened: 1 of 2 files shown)


zhen0 (changes requested):
I think we need to look into this a bit further - I did a quick check on this and although the quick run button is enabled (for a flow with required parameters set as default in the code), I get 'There was a problem creating your run
Error' when I click quick run.  The mutation has a 'Required parameters were not supplied' error.

zangell44:
@znicholasbrown @zhen0 it looks like we're checking against the flow group default parameters when creating the flow run. I'm pretty confused on flow group vs. flow default parameters, maybe we should chat this through a bit today? https://github.com/PrefectHQ/server/blob/master/src/prefect_server/api/runs.py#L163

madkinsz:
Can confirm that this persists when creating runs from the CLI as well.

MRE flow:

```
from prefect import task, Flow, Parameter


@task(log_stdout=True)
def say_data(data):
    print(repr(data))


with Flow("required-default-param") as flow:
    data = Parameter("data", required=True, default=True)
    say_data(data)
```

```
❯ prefect register -p flow.py --project example
❯ prefect run -n required-default-param --execute
Looking up flow metadata... Done
Creating run for flow 'required-default-param'... Error
  Traceback (most recent call last):
    File "/Users/mz/prefect/core/src/prefect/cli/run.py", line 75, in try_error_done
    yield
    File "/Users/mz/prefect/core/src/prefect/cli/run.py", line 639, in run
    flow_run_id = client.create_flow_run(
    File "/Users/mz/prefect/core/src/prefect/client/client.py", line 1412, in create_flow_run
    res = self.graphql(create_mutation, variables=dict(input=inputs))
    File "/Users/mz/prefect/core/src/prefect/client/client.py", line 549, in graphql
    raise ClientError(result["errors"])
  prefect.exceptions.ClientError: [{'path': ['create_flow_run'], 'message': "Required parameters were not supplied: {'data'}", 'extensions': {'code': 'INTERNAL_SERVER_ERROR'}}]
```

zhen0:
Can we close this one @znicholasbrown?

znicholasbrown:
Yeah I think that's fine @zhen0

PR #936, opened 2021-07-07
Improve type handling when passing parameters to create flow run mutation

znicholasbrown:
PR Checklist:

## Describe this PR
Fixes an issue where the editor would incorrectly coerce single-item arrays to scalar values and non-strings to strings in JSON editing mode on the run page.

### DIFF
--- a/src/components/CustomInputs/DictInput.vue
+++ b/src/components/CustomInputs/DictInput.vue
@@ -75,7 +75,14 @@ export default {
           (this.json && Object.keys(JSON.parse(this.jsonInput)).includes(k)) ||
           (!this.json && this.includedKeys.includes(k))
         ) {
-          dict[k] = this.values[i]
+          try {
+            dict[k] =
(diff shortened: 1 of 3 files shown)


src/components/CustomInputs/DictInput.vue
```diff
@@ -182,6 +191,7 @@ export default {
           Object.fromEntries(v.map(entry => [entry.key, entry.value]))
         )
       } else {
+        // console.log()
```
zhen0, line 194: Could we take this out?

zhen0 (changes requested):
One tiny nitpick from me but the type handling improvements look good.

znicholasbrown:
@zangell44 - we haven't traditionally tested these components cause the tests ended up being very brittle but I agree this is something we should test in the future

zhen0 (approved):
Thanks for the change @znicholasbrown

PR #942, opened 2021-07-12
Add run config tab to flow run page

whimsicallyson:
PR Checklist:

## Describe this PR
If a run has custom config changes via the Run tab in the UI, these changes hadn't been shown once the run happens. This adds a tab to the flow run page so you can see what has been changed (or be told that nothing was changed)
Solves #832

### DIFF
--- a/src/pages/FlowRun/Run-Config-Tile.vue
+++ b/src/pages/FlowRun/Run-Config-Tile.vue
@@ -1,9 +1,14 @@
 <script>
 import { runConfigs } from '@/utils/runConfigs'
 import Label from '@/components/Label'
+import JsonInput from '@/components/CustomInputs/JsonInput'
+import YamlInput from '@/components/CustomInputs/YamlInput'
+
 export default {
   components: {
-    Label
+    Label,
+    JsonInput,
+    YamlInput
   },
   props: {
     flowRun: {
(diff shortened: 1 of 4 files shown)


src/pages/FlowRun/Run-Config-Tile.vue
```diff
@@ -0,0 +1,148 @@
...
+        class="row-divider pt-2 pb-8"
+      >
+        <v-row>
+          <v-col cols="4">
+            <v-card-subtitle class="text-h6">{{ arg }}</v-card-subtitle>
+          </v-col>
+          <v-col cols="8">
+            <v-card-text v-if="arg === 'Labels'">
+              <span v-for="(label, j) in argList[arg]" :key="j"
+                ><Label disabled>{{ label }}</Label></span
```
zhen0, line 100: I think we need to add a tiny bit of padding between labels: 

Screen Shot 2021-07-14 at 3 43 55 PM

zhen0 (changes requested):
Thanks for this @whimsicallyson - this is looking nice.  One small request from me.

zhen0 (commented):
Looks good to me.  Any thoughts @znicholasbrown?

znicholasbrown (changes requested):
Screen Shot 2021-07-19 at 11 58 06 AM

Some things I noticed:
- for a default universal config, this screen is mostly whitespace; let's make the height for the container fit the contents
- the bottom few pixels of the container is cut off (notice the box shadow is missing)
- I don't think "Agent" is the correct label for the run config type, let's use "Run Configuration" as we do on the run page
- Lists of image pull secrets shouldn't show index numbers (see screenshot below)

There might be a better way to show plaintext values like these, where there isn't a confusion between keys and values and where templates use the syntax highlighting that we have:
Screen Shot 2021-07-19 at 12 15 27 PM

I'd also reinforce the feedback that values should be more pronounced than keys (like CPU Request, Memory Limit, etc)

whimsicallyson:
@znicholasbrown addressed everything except the yaml output, I haven't been able to find a way to reliably parse the plaintext string. Open to ideas!

znicholasbrown:
You should be able to use the `YamlInput` component to handle that, though with a `disabled` prop

whimsicallyson:
yaml and json are using the input components now! one final concern I can think of: for key/val pairs that can be input as KV or JSON, we have no way to differentiate how they were input, so are all being shown as JSON now (previously they were shown as plaintext KV pairs). I feel like this is ok though?

zhen0:
Is this ready for review @whimsicallyson?

whimsicallyson:
@zhen0 not currently, there was some discussion with nicholas that wasn't captured here and I still need to finish implementing

whimsicallyson:
Closing this (at least for the time being), have some other things that have taken priority. Hope to get back to it soon though!

PR #946, opened 2021-07-15
Setup nginx to run on rootless mode

davzucky:
PR Checklist:

- [x ] add a short description of what's changed to the top of the `CHANGELOG.md`
- [x ] add/update tests (or don't, for reasons explained below)

## Describe this PR

They are more and more people running prefect server on rootless environment on Kubernetes. Some people complain about it on the [slack channel ](https://app.slack.com/client/TL09B008Y/search/search-06c61c24-41a1-40c6-abc6-e8c495c4fccc/thread/C014Z8DPDSR-1623957947.133300).
This change force nginx to run on root less mode

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -4,7 +4,7 @@
 
 ### Features and Improvements
 
-- None
+- Setup Ngnix to run on rootless mode
 
 ### Bugfixes
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -44,6 +44,23 @@ RUN chmod a+x /start_server.sh
 COPY ./intercept.sh /intercept.sh
 RUN chmod a+x /intercept.sh
 
+# Allow write access to www to ensure we can set the configuration file.
+# This is to enable rootless mode
+ARG UID=101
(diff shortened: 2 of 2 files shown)


TylerWanner:
it is great to change to rootless--should we instead just use this base image? https://github.com/nginxinc/docker-nginx-unprivileged

davzucky:
> 
> 
> it is great to change to rootless--should we instead just use this base image? https://github.com/nginxinc/docker-nginx-unprivileged

I don't have any preference. The result would be the same. I know this is what we are using today to customize the base ui image to rootless.

I can change to use the base nginx rootless image if you prefer

znicholasbrown:
I think that makes more sense @davzucky - would you mind changing the base image?

davzucky:
pushed the change to use nginxinc/nginx-unprivileged:stable as a base image

znicholasbrown:
LGTM @davzucky 👍🏻 !

TylerWanner:
thanks @davzucky !

PR #954, opened 2021-07-18
add logrocket session url to support requests

zhen0:
PR Checklist:

### DIFF
--- a/src/pages/Support.vue
+++ b/src/pages/Support.vue
@@ -36,6 +36,16 @@ export default {
     },
     mdScreen() {
       return this.$vuetify?.breakpoint?.mdAndUp
+    },
+    messageWithLogRocketSession() {
+      let url = ''
+      LogRocket.getSessionURL(sessionURL => {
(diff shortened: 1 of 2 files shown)


src/pages/Support.vue
```diff
@@ -36,6 +36,16 @@ export default {
...
+    },
+    messageWithLogRocketSession() {
+      let url = ''
+      LogRocket.getSessionURL(sessionURL => {
+        url = sessionURL
+      })
+      return this.selectedCategory === 'I need help' ||
+        this.selectedCategory === "I've found a bug"
+        ? `${this.message} Relevant LogRocket URL: ${url}`
+        : this.message
```
znicholasbrown, line 48: I think we should wrap this in a try catch in case the LogRocket instance fails for some reason - we don't want to block submission of the form in those cases.

zhen0: Thanks -I'll switch to draft and update next week

PR #957, opened 2021-07-21
Auth exceptions

znicholasbrown:
PR Checklist:

## Describe this PR
This PR updates Okta-auth-js to the latest stable version and adds several deliberate session logs to the start up flow. It also introduces a new worker route (`clear`) for resetting tokens across all webkit-based sessions in cases where we've failed to obtain either id or authorization tokens.


whimsicallyson (dismissed):
This all looks good to me--code makes sense, I was able to log out and back in without issue, logging out in one tab caused sessions in other tabs to log out as well. 

The one thing I'd test with more time is letting a session sit overnight and get stale, as that's sometimes been wonky in the past, but I know you wanted to try and get this out today!

PR #958, opened 2021-07-21
remove jq

TylerWanner:
PR Checklist:

- [x] add a short description of what's changed to the top of the `CHANGELOG.md`
- [x] add/update tests (or don't, for reasons explained below)

## Describe this PR
removes jq dependency
unbreaks the build (failing due to trying to apt-get install in rootless nginx build)

PR #964, opened 2021-07-25
Fix Bash syntax error in start_server.sh

jamesmishra:
PR Checklist:

## Describe this PR

The Prefect UI Docker entrypoint is currently broken because the `start_server.sh` script has a syntax error. Bash expects a `fi` keyword to close out an `if` statement before the end of the file.

This amends a change introduced in PR #958.

You can verify both the bug and the fix by copy-and-pasting this Bash script into https://www.shellcheck.net/.

### DIFF
--- a/start_server.sh
+++ b/start_server.sh
@@ -2,7 +2,18 @@
 
 # Pipes configured environment variables into the 
 # server settings.json file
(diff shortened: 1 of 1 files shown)


TylerWanner:
thanks for fixing! LGTM

znicholasbrown (approved):

PR #974, opened 2021-08-03
Fix start server settings

TylerWanner:
PR Checklist:

## Describe this PR
give non-root nginx user privileges over /var/www in Docker image
institute placeholders and use `sed` to replace them at runtime for serving Server settings at /settings.json

fixes a bug introduced when introducing non-root nginx user and removing jq from the image here https://github.com/PrefectHQ/ui/pull/958


znicholasbrown (approved):
Tested the build locally and confirmed this works against Core 0.15.3 👍🏻

PR #981, opened 2021-08-04
Fix worker error handling

znicholasbrown:
PR Checklist:

- [x] add a short description of what's changed to the top of the `CHANGELOG.md`
- [ ] add/update tests (or don't, for reasons explained below)

## Describe this PR
Fixes an issue where we weren't properly capturing asynchronous errors from the auth worker.

Depends on #980

PR #982, opened 2021-08-04
Fall back to in-memory token exchange if worker exchange fails or takes too long

znicholasbrown:
PR Checklist:

## Describe this PR
Introduces an arg to the `login` method that immediately falls back to in-memory token exchange instead of using the worker. The startup method now uses a promise race with a 10 second timeout to take advantage of this if the worker throws an error on login or takes too long.

Depends on #980 and #981

### DIFF
--- a/src/auth/index.js
+++ b/src/auth/index.js
@@ -62,16 +62,24 @@ if (TokenWorker?.port) {
         }
         break
       case 'console':
+        // eslint-disable-next-line no-console
+        console.log('TokenWorker console', payload)
+        LogRocket.track('TokenWorker console', payload)
+        break
       case 'error':
-        LogRocket.track('TokenWorker Message', payload)
+        // eslint-disable-next-line no-console
+        console.log('TokenWorker error', payload)
+        LogRocket.captureException(payload)
         break
       default:
+        // eslint-disable-next-line no-console
+        console.log('default', payload)
+        LogRocket.track('TokenWorker Message', payload)
         break
     }
   }
 
-  // TODO: Implement error handling from the token worker
-  TokenWorker.port.onmessageerror = e => {
+  TokenWorker.onerror = e => {
     // eslint-disable-next-line no-console
     console.log('Error message received from Token Worker', e)
     LogRocket.captureException(e)
--- a/src/auth/index.js
+++ b/src/auth/index.js
@@ -62,8 +62,9 @@ if (TokenWorker?.port) {
         }
         break
       case 'console':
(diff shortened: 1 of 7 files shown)


src/auth/index.js
```diff
@@ -177,11 +187,13 @@ export const login = async () => {
...
-      source = 'AuthClient'
+      source = 'AuthClient <> InMemory'
       idToken = loginResponse?.idToken
 
-      authorizationTokens = await authorize(idToken.idToken || idToken.value)
-      refresh(authorizationTokens)
+      if (idToken && (idToken.idToken || idToken.value)) {
+        authorizationTokens = await authorize(idToken.idToken || idToken.value)
+        refresh(authorizationTokens)
+      }
```
zhen0, line 196: Should we have a finally block for lines 204 on?

src/main.js
```diff
@@ -84,7 +92,20 @@ export const start = async () => {
...
     // we swap out these tokens later
     loading = true
-    const tokens = await login()
+
+    let tokens
+
+    // We check local storage for a fallback token and compare it against
+    // the curent timestamp - 1 minute. If it falls outside that window
+    // we attempt the normal flow
+    const fallbackTimestamp = localStorage.getItem('prefect_fallback_auth')
```
zhen0, line 101: Are we clearing localStorage anywhere? Would session storage (which clears after each session) be a better fit here? (I guess it doesn't matter as we check date but may slow things down a little to call timeout twice?)

zhen0 (commented):
A few questions from me @znicholasbrown - be good to go over this if you have time tomorrow.

src/auth/index.js
```diff
@@ -62,16 +62,26 @@ if (TokenWorker?.port) {
...
       case 'console':
+        // This should only be used for debugging
+        // eslint-disable-next-line no-console
+        // console.log('TokenWorker console', payload)
+        LogRocket.track('TokenWorker console', payload)
+        break
       case 'error':
-        LogRocket.track('TokenWorker Message', payload)
+        // eslint-disable-next-line no-console
+        console.log('TokenWorker error', payload)
```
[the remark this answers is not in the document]
zhen0: Discussed - yes it is!

src/auth/index.js
zhen0: Not needed (in fact it came up as an unsafe finally block)

zhen0 (approved):
LGTM @znicholasbrown

PR #984, opened 2021-08-05
Update the data store retrieval patterns to handle more than 1000 flows

znicholasbrown:
PR Checklist:

## Describe this PR
Fixes an issue where we were unable to fetch flows from the store because of a cap of 1000 items retrieved from the API. This modifies the `activateFlow` method to instead fetch only 1 flow and insert it into the store instead of attempting to retrieve all of them.


zhen0 (approved):
LGTM @znicholasbrown

PR #987, opened 2021-08-05
Separate primary and secondary Apollo client links

znicholasbrown:
PR Checklist:

## Describe this PR
This PR removes the Batch link from the secondary Apollo client, which is used for non-component-based API requests (like auth, api status, etc). This results in significant improvements in load time (up to 40% faster over the current setup) due to these requests no longer needing to wait for the batch interval to complete. 

I've also increased the batch interval for the primary Apollo client to 2 seconds (up from 1 second) and increased the batch size to 25 (up from 20). Overall this should have very little visible impact on load times but will offset any potential increase in requests from the secondary client.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -4,7 +4,7 @@
 
 ### Features and Improvements
 
-- None
+- Separate primary and secondary Apollo client links - [#987](https://github.com/PrefectHQ/ui/pull/987)
 
 ### Bugfixes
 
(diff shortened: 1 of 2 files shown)


zhen0:
Apologies for review taking a while - as background, I don't think connected to the PR,  but I'm having trouble getting server to run locally when I review this.

zhen0 (approved):
Ok!  Up and running on server and LGTM @znicholasbrown!

Issue #998, opened 2021-08-11
Free Tier Users Have Options to Create Service Account Keys with Restricted Permissions


kvnkho:
**What**

Free tier users see `Read Only` and `User` Roles when creating Service Account Keys in the UI. They then get an Unauthorized message when they try to do this. Instead, they should not be seeing these roles in the drop down in the UI.

* [ ] Ensure that users only see Roles that have restricted permissions on Enterprise accounts. For accounts without RBAC, they should not see these roles.

**Notes**

![image](https://user-images.githubusercontent.com/32503212/129095989-fa557953-8b18-4c4b-b569-4c72fc8167bf.png)
![image](https://user-images.githubusercontent.com/32503212/129095999-abd74bfc-8a8f-4eb7-af3c-cd0d76a6c403.png)

sethschori:
Thanks for opening this bug report @kvnkho!

zhen0:
Closed by #1001

PR #1001, opened 2021-08-12
Service account roles 

ThatGalNatalie:
PR Checklist:

## Describe this PR
#998

### DIFF
--- a/src/pages/TeamSettings/Service-Accounts.vue
+++ b/src/pages/TeamSettings/Service-Accounts.vue
@@ -61,8 +61,13 @@ export default {
     filteredRoles() {
       if (!this.roles) return []
       let rolesToRM = ['RUNNER']
-      if (!this.hasPermission('license', 'admin')) {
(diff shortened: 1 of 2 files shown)


src/pages/TeamSettings/Service-Accounts.vue
```diff
@@ -61,8 +61,13 @@ export default {
     filteredRoles() {
       if (!this.roles) return []
       let rolesToRM = ['RUNNER']
-      if (!this.hasPermission('license', 'admin')) {
```
zhen0, line 64: Sorry I think I gave bad advice when you checked this with me yesterday - we will want to check for both license:admin to see if we should show enterprise license admin and then also check for feature:basic-rbac to see if we should show user and read only (and should remove runner for all).

zhen0 (dismissed):

zhen0 (commented):
Thanks for the changes @ThatGalNatalie - LGTM.  I'll let @znicholasbrown give final approval.

znicholasbrown (approved):
Thanks @ThatGalNatalie 👍🏻

PR #1002, opened 2021-08-16
Expose retry for archived runs

cicdw:
PR Checklist:

## Describe this PR
This PR makes two minor adjustments:

- new runs cannot be created for archived flows; _however_, runs that already exist for archived flows can still be resubmitted / interacted with.  We found some users who used this functionality by restarting flow runs corresponding to archived flows via the UI -- the "restart" button was being exposed on individual task run pages but not on the main flow run page.  This PR now exposes that button for user convenience (**cc:** @dylanbhughes )
- the Core Version deployed with the backend is only relevant information for Prefect Server deploys (as it essentially proxies the version of Server) but is not relevant for Prefect Cloud's UI; this PR hides this info for Cloud only

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -9,6 +9,9 @@
 - Add new login logic to browser visibility handler - [#986](https://github.com/PrefectHQ/ui/pull/986)
 
 ### Bugfixes
+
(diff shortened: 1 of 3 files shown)


znicholasbrown (approved):
LGTM @cicdw 👍🏻

PR #1013, opened 2021-08-24
Remove reliance on private FA npm server

znicholasbrown:
PR Checklist:

## Describe this PR
Removes references to the private FA npm server, adds static assets and updates public/index.html to include static assets that were previously included by the `file-downloader` script.


whimsicallyson:
it won't let me load all the files, but as far as I can tell this all looks good (checked dockerfile, file-downloader, etc)

One thought--is it worth filtering out the things we don't use? Not individual icons, but stuff like the LESS configs, and I want to say we don't use the "solid" icon library (and possibly others)

znicholasbrown:
I think having them available is probably fine cause they won't be requested unless they're needed by the application (that's what the FA script is meant to do so you don't have to load the entire stylesheet). If this leads to a significant blow up in the docker image size I'd be willing to revisit but it'll take awhile to go through all the files and remove the ones we may not need since we don't know everything the script is calling when loading icons.

PR #1019, opened 2021-08-31
add set default role to created instead of on apollo query

zhen0:
PR Checklist:

### DIFF
--- a/src/pages/TeamSettings/Members.vue
+++ b/src/pages/TeamSettings/Members.vue
@@ -220,6 +220,9 @@ export default {
       })
     }
   },
+  created() {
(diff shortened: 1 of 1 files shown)


src/pages/TeamSettings/Members.vue
```diff
@@ -230,7 +235,6 @@ export default {
       pollInterval: 10000,
```
zhen0, line 235: I think the best fix is to move the where we set the role but could/should we also extend this poll interval?  Can't remember if there was a reason we need to be polling so frequently here?

znicholasbrown: I think we had this at 60 s previously but it was leading to some confusion when joining teams/making updates.

znicholasbrown (approved):

PR #1026, opened 2021-09-01
Remove global max width

znicholasbrown:
PR Checklist:

## Describe this PR
This PR removes the global max width of 1440px on the critical path (dashboard > flow > flow run > task > task run) to make better use of large monitors. Other pages are untouched at the moment.

Resolves: #1016


joeclayallday:
Here's another approach if we want to save space. I think your approach makes a ton of sense too & its simpler.

Show-Full-Name

zhen0 (approved):

PR #1028, opened 2021-09-01
Failed flows card should use start_time timestamp instead of updated

znicholasbrown:
PR Checklist:

## Describe this PR
Switches the filter timestamp on the dashboard failed flow tile.


znicholasbrown:
Ok it looks like that's not an indexed field, so I've updated this to use `scheduled_start_time`, which is index and should capture those runs that don't have `start_time` populated.

zhen0 (approved):
Thanks @znicholasbrown - LGTM

PR #1031, opened 2021-09-01
Expand quick runnable flows

znicholasbrown:
PR Checklist:

## Describe this PR
This PR enables quick running flows whose required parameters have defaults set in code but not at the flow group level.


zhen0:
Same issue as #935  - the problem is not the UI Quick Run button - running a flow with required parameters throws an error even with a default provided.  I'm actually not sure why you would add "required=True" at code level if you were also setting a default - they pretty much cancel each other out - so I'm not quite sure this one's even an issue?

Issue #969, opened 2021-07-29
Add copy to task run result location


joshua-gould:
![screenshot](https://user-images.githubusercontent.com/5237151/127522420-adf5183d-c1f1-4e8f-a8da-c333fc7562f4.png)
Since task result locations can be long and are not fully displayed, it would be helpful to be able to copy the value.

PR #1033, opened 2021-09-01
Add a copy method to task run result locations

znicholasbrown:
PR Checklist:

## Describe this PR
Adds a copy function to task run result locations.

Resolves: #969


src/pages/TaskRun/Details-Tile.vue
```diff
@@ -227,7 +243,23 @@ export default {
...
+                      class="cursor-pointer show-icon-hover-focus-only pa-2px"
+                      role="button"
+                      @click="
+                        copyTextToClipboard(
+                          taskRun.serialized_state._result.location
+                        )
+                      "
+                      v-on="on"
+                    >
+                      <v-icon x-small class="mb-2px mr-2" tabindex="0">
```
zhen0, line 256: Should we add a v-if here to check if there's actually a result location to copy?

zhen0 (commented):
One small suggestion @znicholasbrown otherwise LGTM

src/pages/TaskRun/Details-Tile.vue
znicholasbrown: Yeah that's sensible - the copy method won't throw an error since it checks for text to copy but I like hiding the icon in the case as well.

znicholasbrown:
Updated 👍🏻

zhen0 (approved):
Thanks @znicholasbrown

PR #1038, opened 2021-09-02
pagerduty integration 

kingsleyb:
PR Checklist:

## Describe this PR

WIP PR for pagerduty integration

ToDo:

- [x] 1. create a list of keys form integration_keys
- [x] 2. iterate over that list for each row
- [x] 3. add method to be able to delete from list to be saved
- [x] 4. activate save when ready
- [x] 5. fix undefined in action name to be saved
- [x] 5. Add a header explaining the page
- [x] 6. Add a trigger button with a test trigger


zhen0:
I've added most of the To Do list from above.  To both simplify the pager duty page and to add functionality for other action types, I added an actions page in the settings that incorporates the delete action and test action methods.  Main outstanding To Dos are design and wording related.
- What text do we want to in the SOME TEXT HERE space?
- Instructions for Pager Duty integration
- Once all actions are created what do we want to show? I put in a placeholder with links to the Actions page and the dashboard.  



Screen Shot 2021-09-07 at 12 40 13 PM
Screen Shot 2021-09-07 at 12 39 57 PM
Screen Shot 2021-09-07 at 12 39 42 PM

zhen0:
Quick note - you can ignore the Props/IntegrationKeys/Account details printed out at the top of the page.  I think those are for dev purposes only.

src/components/Nav/TeamMenu.vue
```diff
@@ -102,6 +102,20 @@ export default {
           </v-list-item-content>
         </v-list-item>
 
+        <v-list-item :to="'/team/actions'">
+          <v-list-item-avatar>
+            <span style="color: Dodgerblue;">
```
whimsicallyson, line 107: I think we want to move this to the CSS and have it be Prefect blue

whimsicallyson:
Going to let @kingsleyb or someone else weigh in on the descriptive text for the initial page. 
- We should definitely have a header on there, though, and some descriptive text. Do we also want to be able to add another row of action data, or will that all be populated by the pagerduty side of things?
- For the Action Created page, I would expect to see data of the action, as a confirmation that the correct info went through?
- For the actions page in settings, I'm not sure a simple table is the best option, particularly for the config object. Could we set it up more like the K/V pairs in the run config, or have a dropdown like in the K/V Store page?
- I also think the action buttons are given too much space in the settings table, they could be condensed to the right

kingsleyb:
For the header, we should have some pager duty branding and colors, see link here and a sentence that shows the purpose of creating an action.  PagerDuty branding is here https://www.pagerduty.com/brand/

Once the actions are saved, a link to the automations so the actions can be added makes sense, but happy for input on the best experience here.

zhen0:
I've added Pager Duty logo and had a first attempt at wording for the page.  Also added a drop down for the config and included on the action created page - thanks for that feedback @whimsicallyson. 

Screen Shot 2021-09-13 at 11 27 50 AM
Screen Shot 2021-09-13 at 11 28 09 AM
Screen Shot 2021-09-13 at 11 28 32 AM

whimsicallyson:
I have a slight preference towards centering the title, info and save button on the PagerDuty screen, as well as including some padding (basically make it more like the Automations Actions page or others in the Team menu).

I also think we should condense the right-hand buttons on the Actions page, make the test button blue, and move the delete icon to the far right. That'll keep it more in line with other pages of the same style (like Team Members)

zhen0:
Updates made @whimsicallyson and @znicholasbrown - ready for review. 

Note on the pagerduty integration page - since this is an integration page and not part of team settings, I've not centered the header.  I tried it but ended up with a lot of white space. 

I've updated the actions page table to make the action icons closer together and shift delete to the right. 

I believe this will be broken in staging until https://github.com/PrefectHQ/cloud/pull/3730 is merged as we're removing the requirement to add an api secret.

zhen0:
Screen Shot 2021-09-16 at 10 17 11 AM
Screen Shot 2021-09-16 at 10 24 08 AM

whimsicallyson (dismissed):
everything I see has been addressed! 👍

src/pages/TeamSettings/Actions.vue
```diff
@@ -0,0 +1,361 @@
...
+                :dict="item.action_config"
+                :rules="[1, 2]"
+              />
+            </td>
+            <td v-else :colspan="headers.length">
+              No action config
+            </td>
+          </template>
+
+          <!-- TEST ACTION --->
```
znicholasbrown, line 285: Can this be removed?

src/plugins/vuetify.js
```diff
@@ -8,6 +8,7 @@ import NomadIcon from '@/components/Icons/Nomad.vue'
 import PrefectIcon from '@/components/Icons/Prefect.vue'
 import TwilioIcon from '@/components/Icons/twilio.vue'
 import PagerDutyIcon from '@/components/Icons/PagerDuty.vue'
+import newPagerDutyIcon from '@/components/Icons/newPagerDuty.vue'
```
znicholasbrown, line 11: Is there a reason not to replace the existing pager duty icon?

src/router.js
```diff
@@ -242,6 +250,20 @@ export const routes = [
...
       import(/* webpackChunkName: "agent" */ '@/pages/Agents/AgentPage.vue')
   },
+  // ---------------------------//
+  //
+  // Integrations
+  //
+  // ----------------------------//
+  {
+    name: 'pagerduty',
+    path: '/:tenant?/pagerduty/:config?',
```
znicholasbrown, line 260: Without a config string, this is just a blank page. Might we add a redirect if that config isn't present?

znicholasbrown (dismissed):
On the action screen, it looks like we've removed the input for pager-duty api token but the other inputs are still right-aligned and we still have a blurb about storing that in a secret, is that intentional?

A few questions @zhen0 @kingsleyb

src/plugins/vuetify.js
zhen0: Yes - we're using the original pagerDuty logo in automations and cloud hooks.  It is a small PD.  This one is the full Pager Duty name we use as a header at the top of the pager duty integrations page.

zhen0:
Screen Shot 2021-09-17 at 9 54 58 AM
>

znicholasbrown (approved):
Thanks for the changes @zhen0 👍🏻

PR #1047, opened 2021-09-12
add run config tab in flow run details

gmaliar:
PR Checklist:

## Describe this PR
We're using a machine learning flow with multiple run configuration environment and image settings so it's hard for us to debug which image / env variables we ran in each run right now as we don't have visual access to the `RunConfig` object

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -4,7 +4,7 @@
 
 ### Features and Improvements
 
-- None
+- Add RunConfig tab in FlowRun details - [#1047](https://github.com/PrefectHQ/ui/pull/1047)
 
 ### Bugfixes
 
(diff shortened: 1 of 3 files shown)


znicholasbrown:
@gmaliar thanks for the PR! It looks like you'll need to add `|| hasRunConfig` to line 91 of the flow run page details tile or this will only show up when a run also has parameters/context

gmaliar:
Ooh good catch, I didn't notice that! 🙇 
Fixed!

znicholasbrown (approved):
This LGTM @gmaliar, thanks!!

gmaliar:
🙏

PR #1064, opened 2021-09-28
update tabs on tiles

stackoverfloweth:
PR Checklist:


src/pages/Flow/UpcomingRuns-Tile.vue
```diff
@@ -41,86 +36,82 @@ export default {
...
-      if (this.lateRuns?.length > 0) {
-        this.tab = 'late'
-      }
-      if (this.lateRuns?.length <= 0) {
-        this.tab = 'upcoming'
+    upcomingFlowRunsData(val) {
+      if (!val || !this.pristine) return
+
+      if (this.tab == this.tabs.upcoming && this.lateRuns?.length > 0) {
+        this.tab = this.tabs.late
```
stackoverfloweth, line 113: quick note here, I was frustrated that if there are runs in the "late" tab, even when I explicitly go back to the "upcoming" tab the system would continue to kick me back to "late" on a regular interval. I change this behavior to only programmatically switch the user to "late" tab if the user hasn't explicitly changed tabs

src/pages/Flow/UpcomingRuns-Tile.vue
```diff
@@ -497,7 +429,7 @@ a {
 
 .card-content {
   height: 100%;
-  max-height: 226px;
```
stackoverfloweth, line 500: this is the height of v-tabs

src/pages/Flow/Details-Tile.vue
```diff
@@ -566,7 +552,7 @@ export default {
 }
 
 .card-content {
-  max-height: 254px;
```
stackoverfloweth, line 569: this is the height of v-tabs

src/pages/Agents/SubmittableRuns.vue
```diff
@@ -93,54 +100,69 @@ export default {
...
-        this.tab = 'late'
-      }
-      if (val.lateRuns?.length <= 0) {
-        this.tab = 'submittable'
+      if (!val || this.userSetTab) return
+
+      if (this.tab == this.tabs.submittable && this.lateRuns?.length > 0) {
+        this.tab = this.tabs.late
+      } else if (this.tab == this.tabs.late && this.lateRuns?.length == 0) {
+        this.tab = this.tabs.submittable
```
stackoverfloweth, line 165: quick note here, I was frustrated that if there are runs in the "late" tab, even when I explicitly go back to the "submittable" tab the system would continue to kick me back to "late" on a regular interval. I change this behavior to only programmatically switch the tab if the user hasn't explicitly changed tabs (same as UpcomingRunsTile)

src/pages/Flow/UpcomingRuns-Tile.vue
```diff
@@ -178,8 +190,13 @@ export default {
       },
       loadingKey: 'loadingKey',
       pollInterval: 10000,
-      fetchPolicy: 'no-cache',
-      update: data => data?.flow_run || []
+      update: data => {
+        return (data?.flow_run || []).map(run => {
+          run.cacheInvalidation = new Date().getTime()
```
pleek91, line 195: nitpick: make `new Date().getTime()` a const and use the same date for every flow run.

src/pages/FlowRun/Details-Tile.vue
```diff
@@ -343,41 +276,29 @@ export default {
...
+      </v-tab-item>
+    </v-tabs-items>
   </v-card>
 </template>
 
-<style lang="scss" scoped>
+<style lang="scss">
 .card-content {
-  max-height: 254px;
+  max-height: calc(254px - var(--v-tabs-height));
```
pleek91, line 301: I think since your `v-tabs` has a `v-if` on it you need to conditionally subtract `--v-tabs-height`

pleek91 (approved):
Handful of tiny suggestions. Looks great overall

zhen0 (changes requested):
This looks nice @stackoverfloweth.  A few small comments from me (mainly about var vs const).  Also - could we include the upcoming/late runs tile on the dashboard in these changes?  It feels better to have that switch across the board.

src/pages/FlowRun/Details-Tile.vue
stackoverfloweth: good catch, although upon closer inspection it seems that class isn't actually used in this template

src/pages/Flow/UpcomingRuns-Tile.vue
stackoverfloweth: 🧠

zhen0 (commented):
One tiny nitpick - there's a border above 'options' on the late run tab that I don't think we want/need. it's especially obvious in dark mode. 

This PR:

Screen Shot 2021-09-30 at 5 11 23 PM

Screen Shot 2021-09-30 at 5 11 36 PM

Current UI:

Screen Shot 2021-09-30 at 5 12 19 PM
Screen Shot 2021-09-30 at 5 12 32 PM

zhen0 (approved):

zhen0:
Looks good to me - can we get another approval before merging to staging?  @pleek91 @whimsicallyson

whimsicallyson:
my only question/comment is how it works if there are more tab options than the width of the card. does it overflow properly in the horizontal, or do we wend up with the same issue that we had with the old implementation?

stackoverfloweth:
> my only question/comment is how it works if there are more tab options than the width of the card. does it overflow properly in the horizontal, or do we wend up with the same issue that we had with the old implementation?

@whimsicallyson Great point, yes it does overflow properly! That's built into v-tabs https://vuetifyjs.com/en/components/tabs/#dynamic-tabs.
You can see an example of this on the tile reference in the original bug (can have up to 4 tabs)

whimsicallyson (approved):

PR #1070, opened 2021-09-30
update pagerduty logo color in dark mode

zhen0:
PR Checklist:


zhen0:
Hmmm... You're correct that it's visible in dark mode but the change was based on PagerDuty brand rules that the PagerDuty green should only be used on a white background. I'll see if we can get more info about how strictly they apply that and then we can close or adjust as needed.

PR #1096, opened 2021-10-26
when creating/editing schedule, make action button available across tabs

### LINKED ISSUE #1061: when creating/editing schedule, make action button available across tabs
**Why**
the action of "Create" (schedule) or in the future "Save" (edited schedule) is only available on the "schedule" tab, so when a user is on the "parameters" tab it is unclear if an action is required or if save happened implicitly.  

**What**
the action should be available whenever editing a schedule, regardless of what tab is selected (currently only "schedule" and "parameters" tabs)

**Notes**
Screen Shot 2021-09-27 at 4 41 15 PM
Screen Shot 2021-09-27 at 4 40 48 PM

ThatGalNatalie:
PR Checklist:

## Describe this PR
Should resolve #1061 

This moves the `DictInput` component and its computed prop and methods from the `Schedules` component into the `ClockForm` component

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -4,6 +4,7 @@
 
 ### Features and Improvements
 
+- When creating/editing schedule, make action button available across tabs - [#1096](https://github.com/PrefectHQ/ui/pull/1096)
 - Add warning that users will be locked out if not part of a team - [#1078](https://github.com/PrefectHQ/ui/pull/1078)
 - Update the PagerDuty logo to match new branding - [#1073](https://github.com/PrefectHQ/ui/pull/1073)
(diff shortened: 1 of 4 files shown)


zhen0 (dismissed):
Thanks for the changes @ThatGalNatalie.  Can you make sure master is updated?  I'd also like to co-ordinate this with #1104 - can you check in with @stackoverfloweth to make sure the two don't conflict too much.

ThatGalNatalie:
Since #1104 depends on #1095 which will most likely get merged later on this week, Evan has said it's okay to merge this pr first

PR #1101, opened 2021-10-29
1095/team settings

stackoverfloweth:


zhen0 (dismissed):
Looks like there's a stray console log in here somewhere.  

Also spacing needs some tweaks - 
What I see locally:
Screen Shot 2021-11-03 at 12 53 56 PM

What I see in cloud.prefect.io:
Screen Shot 2021-11-03 at 12 53 28 PM

zhen0 (dismissed):
Not had a chance to check the KV store (need to switch backend/upgrade to find a tenant with KV access).  Actions sections look fine but looks like there's an issue setting secrets.

src/pages/TeamSettings/KV.vue
```diff
@@ -218,28 +207,12 @@ export default {
           return
         }
       }
-      let value = this.KvValueInput
```
zhen0, line 221: Tested KV store using Netlify and see a similar error to the secrets - need error handling (that is in the original version but looks like it's not working here) and to be able to handle a string that is incorrect json (since the expectation is that both strings and json could be added here): 





Screen Shot 2021-11-13 at 11 06 12 AM


Screen Shot 2021-11-13 at 10 55 02 AM

Screen Shot 2021-11-13 at 11 08 10 AM

stackoverfloweth: I believe these concerns should be addressed with latest commits, thanks for the great suggestions

zhen0 (approved):
Thanks for the changes @stackoverfloweth - LGTM!

PR #1103, opened 2021-10-29
Performance: Update inputs in flow settings - parameters

stackoverfloweth:


zhen0:
I'm seeing the same issue here with type not resetting as mentioned on #1095. I think if we're going to show type we need to fully reset the type when reset is clicked.

zhen0:
For info - I'm going to hold off on reviewing this one further until changes in #1095 are made and merged so I'm not commenting on the same things across PRs.

zhen0 (approved):
LGTM @stackoverfloweth

PR #1104, opened 2021-10-29
Performance: Update inputs in flow settings - schedules

stackoverfloweth:


zhen0:
For info - I'm going to hold off on reviewing this one further until changes in #1095 are made and merged so I'm not commenting on the same things across PRs.

ThatGalNatalie:
With #1061 merged, I haven't seen any breaking changes with that but I noticed that pasting in a value in the json editor and then switching back to the dict tab does not reflect my changes

https://user-images.githubusercontent.com/40726391/140410520-69ef62d9-3f91-4acc-83db-555958d25ae2.mov

stackoverfloweth:
> With #1096 merged, I haven't seen any breaking changes with that but I noticed that pasting in a value in the json editor and then switching back to the dict tab does not reflect my changes
> 
>  reflection.mov

Great find @ThatGalNatalie! I resolved this issue with a slight change to the developer experience for checkbox dictionary inputs. The user experience also changed slightly if you want to check it out.

Thanks again!

zhen0 (dismissed):
There's some complexity in how we pass parameters to schedules that need to be factored in for this one.  We allow a user to select which of their parameters should be added as a schedule parameter (and thus is different from the default) and which should use the default.  The checkboxes on the dictinput section allow a user to select which params to include.  The main thing I'm seeing in this branch is that I can make changes in json that are not reflected when I set the schedule because they have not been selected as schedule params on the dict input page.  In current cloud, anytime a param is altered in json, it is updated to be a scheduled parameter.  Happy to chat this through to make sure we're on the same page about the best UX.

zhen0 (approved):
Thanks for the updates @stackoverfloweth LGTM!

PR #1105, opened 2021-10-30
fix ordering of submitted and running

zhen0:
PR Checklist:

### DIFF
--- a/src/pages/Dashboard/InProgress-Tile.vue
+++ b/src/pages/Dashboard/InProgress-Tile.vue
@@ -91,6 +91,7 @@ export default {
       return color
     },
     title() {
+      console.log('tabs', this.tabs, this.tab)
       let title = this.loading
         ? 'In progress flow runs'
         : `${this.all?.length || 0} runs in progress`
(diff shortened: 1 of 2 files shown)


src/pages/Dashboard/InProgress-Tile.vue
```diff
@@ -224,12 +224,12 @@ export default {
...
       </v-tab>
-      <v-tab :key="tabs.running" data-cy="in-progress-tile-running">
-        Running
-      </v-tab>
       <v-tab :key="tabs.submitted" data-cy="in-progress-tile-submitted">
         Submitted
       </v-tab>
+      <v-tab :key="tabs.running" data-cy="in-progress-tile-running">
+        Running
+      </v-tab>
```
stackoverfloweth, line 232: this definitely does solve the problem, but it appears that historically the "running" tab was always listed before the "submitted" tab, this would change that. 

you can solve the users issue without reordering the tabs by changing the value of the tabs at the top (line 37 & 38) so that running has a value of `1` and submitted has a value of `2`. 

I don't feel strongly either way

stackoverfloweth (dismissed):

src/pages/Dashboard/InProgress-Tile.vue
zhen0: Oh interesting.  I deliberately did it this way as I felt that submitted comes before running in a standard flow.  But makes sense to stick with what we had historically for user consistency.  I'll switch them round.

cicdw (approved):

PR #1107, opened 2021-11-03
tree shaking lodash dependency

stackoverfloweth:
PR Checklist:

## Describe this PR
Per Lodash [documentation](https://lodash.com/per-method-packages)

> Lodash methods are available in standalone per method packages like lodash.mapvalues, lodash.pickby, etc. These packages contain only the code the method depends on.
>However, use of these packages is discouraged and they will be removed in v5.

with moving away from per-method-packages and webpacks built in tree shaking, I was able to reduce lodash's impact on our bundle from 531.3K to 100.5K

should have zero impact on users


zhen0 (dismissed):
LGTM @stackoverfloweth - would you mind updating to the latest version master so we can merge?  Thanks!

znicholasbrown (approved):

znicholasbrown (approved):

PR #1108, opened 2021-11-03
tree shaking highlightjs dependency

stackoverfloweth:
PR Checklist:

## Describe this PR
This package's use of Highlight.js was over double the size of the next biggest dependency. Removing this dependency and the 3 lines from markdownParser I was able to bring down the highlightjs bundle size from 1.3M to 77.5K! 

While at first glance, I expected this to have an impact on user experience, I cannot find an instance where this change actually effected what the user sees. I confirmed the "describe" component used in a flow's readme, as well as the artifacts tab of a flow-run. Markdown entry and parsing seems uneffected by this change.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -9,6 +9,7 @@
 
 ### Bugfixes
 
+- Removed unnecessary dependency preventing us from tree shaking highlightjs - [#1108](https://github.com/PrefectHQ/ui/pull/1108) 
 - Adds a try/catch block when deleting a membership in the removeUser method - [#1094](https://github.com/PrefectHQ/ui/pull/1094)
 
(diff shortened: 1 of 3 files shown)


zhen0 (dismissed):
This looks good to me @stackoverfloweth.  I also checked tutorials as I believe they also use markdown and looked fine.  As she did some work of the initial work on the markdown parser I'd like to have @ThatGalNatalie take a look at this before we merge.

zhen0 (changes requested):
Oh actually I do have a requested change @stackoverfloweth - can you update the changelog please?  (And update to latest master!). Thanks!

ThatGalNatalie (dismissed):
looks good to me too

zhen0 (approved):
Also looks good to me @stackoverfloweth - but can you add the latest from master so we can merge.

znicholasbrown (approved):

PR #1109, opened 2021-11-03
tree shaking moment dependency

stackoverfloweth:
PR Checklist:

## Describe this PR
using webpacks tooling, I was able to slim down the dependency of moment/locale down to just a handful of locales that we might need in the future (would be easy to adjust or add more when needed). This change brought the moment bundle from 672.3K to 200.1K.

should have no impact on users


vue.config.js
```diff
@@ -93,6 +100,7 @@ module.exports = {
     },
     webpackBundleAnalyzer: {
       analyzerMode: 'disabled'
+      //generateStatsFile: true
```
stackoverfloweth, line 103: I would like to leave this comment here if it's okay with everybody else? This line means that a stats.json file gets generated, which is what I used at https://chrisbateman.github.io/webpack-visualizer/ to visualize the node-modules in our bundle. Might be useful again in the future?

znicholasbrown: This is fine to leave in but please add a comment to the code about why this is commented out 👍🏻

znicholasbrown (dismissed):

znicholasbrown (approved):

PR #1120, opened 2021-11-09
Performance: Add virtual-scroll to dashboard tiles and improve duration calculations

stackoverfloweth:
PR Checklist:

## Describe this PR
There are several improvements made to the dashboard/agents/flow pages with this PR. Most importantly, I'm using v-virtual-scroll in lieu of v-list in 4 important tiles. These tiles caused a significant number of nodes to be created regardless of if the user would scroll through them or not. There will be a slightly worse user experience when scrolling, as components are dynamically added/removed but certainly worth the performance gains IMO. Also, while investigating sub-dependencies of these tile-rows, I found the DurationSpan component, which generated several moment instances every second. I was able to apply the same logic from the CurrentTime component to move that interval to 60 seconds as well as remove the moment dependency from DurationSpan entirely


src/utils/dateTime.js
```diff
@@ -122,3 +127,24 @@ export const roundedOneAgo = unitOftime => {
     .startOf(roundedTo)
     .format()
 }
+
+export function getMillisecondsUntilNextMinute() {
+  return 60000 - new Date().getSeconds() * 1000 + new Date().getMilliseconds()
+}
+
+export function durationDifference(start, end) {
```
pleek91, line 135: this is a cool utility method. nice job

pleek91 (commented):
LGTM

src/utils/dateTime.js
stackoverfloweth: gracias 🙏

src/components/DurationSpan.vue
```diff
@@ -22,37 +23,65 @@ export default {
...
     }
   }
 }
 </script>
 
 <template>
-  <span>{{ duration | duration }}</span>
+  <span class="duration-span">
+    <span class="duration-span__days"
+      >{{ duration.days }} {{ plural('day', duration.days.length) }}</span
```
zangell44, line 66: I think we need to check if `duration` is null for this span, I'm seeing errors locally 

```
Vue Global Error Handler {error: TypeError: Cannot read properties of null (reading 'days')
```

zhen0 (dismissed):
This is looking nice @stackoverfloweth - I can see a reduction in errors in the dev console on this branch. 

One requested change for duration:

If the duration is less than a certain unit (e.g. day) can we exclude that unit (as we currently do in staging):

This branch:

Screen Shot 2021-11-11 at 4 06 54 PM

Desired/current staging:
Screen Shot 2021-11-11 at 4 07 02 PM

And with the latest change there's possibly also a spacing/formatting issue (space before the comma):
Screen Shot 2021-11-11 at 5 00 55 PM

zhen0 (dismissed):
Still going through this one but wanted to note that i'm seeing an issue with running flows not showing in the runs in progress tile:

Screen Shot 2021-11-13 at 11 33 50 AM

zhen0 (dismissed):
When flow runs are in a cancelling state I see the following error message in the console: 
```
"Invalid prop: type check failed for prop \"endTime\". Expected String with value \"Thu Nov 18 2021 17:57:37 GMT-0500 (Eastern Standard Time)\", got Date "
```

This also seems to **very briefly** take up a lot of CPU:

Screen Shot 2021-11-18 at 5 57 46 PM

Otherwise, I can now see all running and submitted states and I've not experienced any page freezes even when running late runs etc.

stackoverfloweth:
> When flow runs are in a cancelling state I see the following error message in the console:
> 
> ```
> "Invalid prop: type check failed for prop \"endTime\". Expected String with value \"Thu Nov 18 2021 17:57:37 GMT-0500 (Eastern Standard Time)\", got Date "
> ```
> 
> This also seems to **very briefly** take up a lot of CPU:
> 
> Screen Shot 2021-11-18 at 5 57 46 PM
> 
> Otherwise, I can now see all running and submitted states and I've not experienced any page freezes even when running late runs etc.

I think this error did exist before me, the date coming down from the graphql query has a native date for some reason, and the component expects either null or string.
Screen Shot 2021-11-29 at 3 54 54 PM

I think the reason we’re not seeing it on dev is maybe because it’s running “production” code via npm run build? I believe that would remove these vue global warnings

I checked out master and confirmed that I get those same warnings on “cancelling” flows

stackoverfloweth:
note from discussion around native `end_time`. I verified that the data coming back from apollo goes from state "scheduled" 
```
{
    "id": "e66d3237-66b0-4065-adf6-0f7d41bc05d2",
    "name": "beryl-raccoon",
    "state": "Scheduled",
    "labels": [
        "Evans-MacBook-Pro-2.local"
    ],
    "scheduled_start_time": "2021-11-30T10:01:33.24699+00:00",
    "version": 1,
    "flow": {
        "id": "7b881aa0-5b0e-49e9-a4d9-f074960f901f",
        "name": "slow-flow",
        "schedule": null,
        "environment": null,
        "is_schedule_active": true
    }
}
```

to state "cancelling", always with a normal looking `start_time` and no `end_time`
```
{
    "id": "742dd5ce-25d6-4c0a-9a1a-7dba31dbf7e1",
    "name": "cinnamon-raccoon",
    "start_time": "2021-11-30T14:39:53.986704+00:00",
    "state": "Cancelling",
    "state_timestamp": "2021-11-30T14:41:41.458238+00:00",
    "flow": {
        "id": "7b881aa0-5b0e-49e9-a4d9-f074960f901f",
        "name": "slow-flow",
        "flow_group_id": "b6856226-145d-4955-a066-e022adb88f04"
    },
    "updated": "2021-11-30T14:41:41.527171+00:00"
}
```

stackoverfloweth:
I found it, this component `Dashboard/FlowRunHistory-Tile.vue` has the apollo query for the data above, passes the items down to `components/TimelineTooltip.vue`, which is ultimately where DurationSpan is throwing these warnings for receiving a native `endTime` instead of null | string. The items passed come from the FlowRunHistory-Tile's mixin `flowRunHistoryMixin.js`, which is setting `end_time` to a new Date() on line 119.

Screen Shot 2021-11-30 at 8 59 32 AM
.

src/components/DurationSpan.vue
```diff
@@ -1,5 +1,16 @@
...
+  toDurationDifferenceString
+} from '@/utils/dateTime'
+import {
+  MS_PER_SECOND,
+  MS_PER_MINUTE,
+  MS_PER_HOUR,
+  getMillisecondsUntilNextSecond,
+  getMillisecondsUntilNextMinute,
+  getMillisecondsUntilNextHour
+} from '@/utils/dateTime'
```
pleek91, line 13: nit: two import statements from the same file.

src/components/DurationSpan.vue
```diff
@@ -16,43 +27,87 @@ export default {
...
+    {{ durationDifferenceString }}
+  </span>
 </template>
+
+<style lang="scss" scoped>
+.duration-span__block:first-child {
+  margin-right: 2px;
+  content: ',';
+}
+</style>
```
pleek91, line 113: What does this css do? There's not class `duration-span__block` in the template.

zhen0:
Thanks for tracking that down @stackoverfloweth - this LGTM and happy to approve once Craig's comments are addressed.

zhen0 (changes requested):
Thanks for tracking that down @stackoverfloweth - this LGTM and happy to approve once Craig's comments are addressed.

zhen0 (approved):
Thanks for the changes @stackoverfloweth

Issue #1080, opened 2021-10-11
Allow users to customize which columns are visible 


stackoverfloweth:
**Why**
User from community reported frustration on the Flow > Runs page specifically. Currently we show name, start, end, and state columns inside of a `v-data-table`. The user specifically is requesting the ability to resize or toggle column visibility. However, the vuetify component comes with some decent support for controlling what columns should be visible on a mobile view vs a desktop view, and at what px width should it break between the two. Vuetify doesn't seem to have support for resizing columns out of the box. 

**What**
- [ ] Investigate the several uses of `v-data-table` throughout the UI project
- [ ] Set appropriate mobile-breakpoints based on the number of columns and average width of content in each. https://vuetifyjs.com/en/api/v-data-table/#props-mobile-breakpoint.
- [ ] Consider also hiding columns that are not critical as the user shrinks down towards the mobile-breakpoint. 
- [ ] Validate that the user experience of tabular data is acceptable on various screen sizes including mobile

**Notes**
https://prefect-community.slack.com/archives/C0192RWGJQH/p1633799892324000

zhen0:
Closed by #1121

PR #1121, opened 2021-11-09
Feature: Adjust column width in v-data-table to account for longer names

ThatGalNatalie:
PR Checklist:

## Describe this PR
#1080 
Adjusts the width to provide more room for the name column while also hiding specific columns depending on the breakpoint. While searching to see where we use v-data-table, I thought that some tables didn't need adjustments to their width and I just kept them as is. These changes are to tables where I thought it would be nice to have more room.


zhen0 (approved):
LGTM @ThatGalNatalie

PR #1134, opened 2021-11-18
Bugfix: Check for skipped state in task run table to fix endless duration 

### LINKED ISSUE #1110: BUG: SKIPPED states don't have end times so UI counts up duration forever
**Bug Description**

Screen Shot 2021-11-03 at 5 15 57 PM

**Steps To Reproduce**

1. Get a task run to raise a `SKIP` signal
2. Go to the Task Run page in the UI or the mapped runs page
3. Watch the duration count slowly upwards forever

**Browsers Tested:**
* [X] Chrome
* [ ] Firefox
* [ ] Safari
* [ ] Edge
* [ ] IE

----------
**Notes**

ThatGalNatalie:
PR Checklist:

## Describe this PR
Resolves #1110


src/pages/TaskRun/MappedTaskRuns-Tile.vue
```diff
@@ -274,7 +279,13 @@ export default {
           <DurationSpan
             v-if="item.start_time"
             :start-time="item.start_time"
-            :end-time="item.end_time"
+            :end-time="
+              item.end_time
+                ? item.end_time
+                : isFinished(item.state)
+                ? item.start_time
+                : null
```
pleek91, line 287: This ternary is copied, good opportunity to create a utility function.

src/pages/TaskRun/MappedTaskRuns-Tile.vue
```diff
@@ -75,7 +76,11 @@ export default {
       return `%${this.searchTerm}%`
     }
   },
-  methods: {},
+  methods: {
+    isFinished(state) {
+      return FINISHED_STATES.includes(state)
+    }
```
pleek91, line 82: let's make this a utility function in an external file so it can be reused in both components.

pleek91 (dismissed):
Everything looks good. But lets DRY up the logic by creating some utility functions.

zhen0 (dismissed):
This looks good to me but I also see that duration count in the task run details tile.  Could we add your durationCalc method there too?
Screen Shot 2021-11-30 at 1 14 12 PM

src/pages/Task/TaskRunTable-Tile.vue
```diff
@@ -71,7 +72,11 @@ export default {
       return `%${this.searchTerm}%`
     }
   },
-  methods: {},
+  methods: {
+    durationCalc(startTime, endTime, state) {
```
stackoverfloweth, line 76: nit: I would import the method with this full name `calculateDuration`, and rather than using another name here but calling the imported method I would just make it available to the template with

```
methods: {
    calculateDuration
}
```

which should work fine since the arguments do not change

src/pages/TaskRun/MappedTaskRuns-Tile.vue
```diff
@@ -75,7 +76,11 @@ export default {
       return `%${this.searchTerm}%`
     }
   },
-  methods: {},
+  methods: {
+    durationCalc(startTime, endTime, state) {
```
stackoverfloweth, line 80: same suggestion here

src/utils/calculateDuration.js
```diff
@@ -0,0 +1,9 @@
+import { FINISHED_STATES } from '@/utils/states'
+
+function isFinished(state) {
+  return FINISHED_STATES.includes(state)
+}
+
+export function duration(startTime, endTime, state) {
```
stackoverfloweth, line 7: I would recommend either naming the function `calculateDuration`, and/or consider having it be the default export (which would require removing curly brackets where you're importing it)

src/utils/calculateDuration.js
```diff
@@ -0,0 +1,9 @@
...
+  return endTime ? endTime : isFinished(state) ? startTime : null
```
stackoverfloweth, line 8: I would be in favor or more lines over nested ternaries. IMO This can be pretty hard to follow compared to 

```
if (endTime) {
   return endTime
}

if (isFinished(state)) {
    return startTime
}

return null
```

src/utils/calculateDuration.js
pleek91: Generally a utilities file wouldn't have a default export. So I'd go with the more clear name route. 

However I'd suggest this utility belongs in the `@utils/states.js` since it has to do with states. a utility file name `calculateDuration` is probably too specific.

src/utils/calculateDuration.js
pleek91: +1

src/utils/calculateDuration.js
ThatGalNatalie: I think putting it in @utils/states.js makes sense too

zhen0 (approved):
Thanks @ThatGalNatalie

PR #1136, opened 2021-11-19
Bugfix: Set min-width in titlebars to account for shorter names

ThatGalNatalie:
PR Checklist:

## Describe this PR
Sets a min-width of 10vw in the Flow, Flow Run, and Task Run titlebars to not cut off shorter names

### DIFF
--- a/src/pages/TaskRun/TaskRun.vue
+++ b/src/pages/TaskRun/TaskRun.vue
@@ -212,6 +212,7 @@ export default {
         :style="{
           display: $vuetify.breakpoint.smAndDown ? 'inline' : 'block',
           width: taskRunNameLength + 'ch',
+          minWidth: '10vw',
           maxWidth: titleBarMaxWidth
         }"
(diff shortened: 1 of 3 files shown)


src/pages/TaskRun/TaskRun.vue
```diff
@@ -212,6 +212,7 @@ export default {
         :style="{
           display: $vuetify.breakpoint.smAndDown ? 'inline' : 'block',
           width: taskRunNameLength + 'ch',
+          minWidth: '10vw',
```
pleek91, line 215: Can we make a css class for this? That way we're not binding an inline style for something that will never change.

pleek91 (changes requested):
Lets use a css class here. That'll keep it DRY as well as keep us from binding an inline style that will never change.

zhen0 (approved):
LGTM @ThatGalNatalie

PR #1139, opened 2021-11-23
Feature: Show project name in the run history flow run pop up box

### LINKED ISSUE #843: Show project name in the run history flow run pop up box
**Why**

It is possible that there are different projects that happen to have the same flow name. However, in the overview tab of the dashboard, using mouse over to see the details of the flow in the run history does not show the project name. This is not an issue when a particular project is selected, but when "All Projects" have been selected, this can be confusing since different flows with the same name are identified in exactly the same way.

**What**
* In the run history, the pop up box on mouse over for a flow run should show the project name near the flow name and the flow run name, at least when "All Projects" have been selected.

**Notes**
* It could suffice to use a smaller font size like the flow run name.

ThatGalNatalie:
PR Checklist:

## Describe this PR
Resolves #843


src/pages/Dashboard/FlowRunHistory-Tile.vue
```diff
@@ -90,7 +90,11 @@ export default {
       @bar-mouseover="_barMouseover"
     >
       <template v-if="canShowTooltip" slot="tooltip">
-        <TimelineTooltip :tooltip="tooltip" :loading="tooltipLoading" />
+        <TimelineTooltip
+          :tooltip="tooltip"
+          :loading="tooltipLoading"
+          :is-dashboard="$route.name === 'dashboard'"
```
stackoverfloweth, line 96: rather than passing this prop, we could just as easily check the $route from inside the TimeLineTooltip?

ThatGalNatalie: The TimelineTooltip is a functional component so I thought since it was instanceless, we wouldn't have access to `this`?

stackoverfloweth: ah, I didn't catch that. That's fine then, perhaps a different name for the prop? I think it would be better to explain what the functionality of the prop is, like `show-project-name`? What do you think?

ThatGalNatalie: Agree, `show-project-name` is more descriptive. Thanks!

zhen0 (approved):
Thanks for the changes LGTM @ThatGalNatalie

zhen0:
@cicdw - another one that needs a review from you- the gql change does involve joining tables but the change is one that's been asked for several times and does not seem slow anything down.

cicdw (approved):

PR #1143, opened 2021-11-29
Performance: Removed watchers and/or increased watchers specificity

stackoverfloweth:
Watchers can be expensive, especially when the object is a large array, or the watcher is "deep".  Even inexpensive watchers can become create performance issues when you're dealing with a large number of the components within a view.  This PR outright removes many watchers. Where watchers couldn't be removed, I was still able to increase the watchers specificity.   For example, watching `$route` only to be checking a very specific param within the handler. It's better to just watch the specific param by watching `$route.params.tenant` for example. Some watchers were replaced with computed variables to follow vue's suggestion that watchers be used for "side-effects of changing data". There are also some cases where I was able to use the watcher API to only temporarily watch a property instead of continuing to watch for the lifetime of the component. Together these changes should help improve performance of the system and also improve the clarity of the intent of watchers within these components.

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -177,13 +177,9 @@ export default {
         this.$apollo.queries.flows.refresh()
       }
     },
-    async $route(new_route, old_route) {
-      if (
-        new_route?.params?.tenant &&
-        new_route?.params?.tenant !== old_route?.params?.tenant &&
-        this.tenant?.slug !== new_route.params.tenant
-      ) {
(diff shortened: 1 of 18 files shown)


zhen0 (dismissed):
Thanks @stackoverfloweth - one small request for the PR itself (no code change needed) - can you add a sentence or two about the benefits of this change?

zhen0 (approved):
Thanks @stackoverfloweth!

PR #1144, opened 2021-11-30
Feature: Add loading state to RestartDialog

ThatGalNatalie:
PR Checklist:

## Describe this PR
Adds a loading and disabled state when restarting flow runs

### DIFF
--- a/src/pages/FlowRun/Actions.vue
+++ b/src/pages/FlowRun/Actions.vue
@@ -183,7 +183,9 @@ export default {
             depressed
             style="height: 46px;"
             small
-            :disabled="!hasPermission('update', 'run') || !canRestart"
+            :disabled="
+              !hasPermission('update', 'run') || !canRestart || restartDialog
+            "
             color="info"
             @click="handleRestartClick"
           >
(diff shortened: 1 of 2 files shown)


stackoverfloweth:
Screen Shot 2021-11-30 at 9 40 10 AM

src/pages/FlowRun/Actions.vue
```diff
@@ -183,7 +183,9 @@ export default {
             depressed
             style="height: 46px;"
             small
-            :disabled="!hasPermission('update', 'run') || !canRestart"
+            :disabled="
+              !hasPermission('update', 'run') || !canRestart || restartDialog
```
stackoverfloweth, line 187: is this change necessary? that dialog is a modal, so the user shouldn't be able to click anything behind the modal when `restartDialog` is true

stackoverfloweth: although, visually I do like the effect

stackoverfloweth (approved):
LGTM

Great addition!
It would be nice if when the restart-dialog closes, the flow-run behind should appear in a "scheduled" or "running" state. Immediately after closing restarting, the flow-run behind still appears cancelled, and I can go through the restart dialog again. I'm not suggesting this should be done as part of your PR.

zhen0:
Added https://github.com/PrefectHQ/ui/issues/1147 to capture Evan's suggestion here.

PR #1149, opened 2021-12-07
Make request credentials forwarding customizable

jverce:
PR Checklist:

## Describe this PR

When running Prefect on premise, it's sometimes necessary for the UI to
forward authentication cookies (e.g. when the authentication layer is provided
by a reverse proxy and not the Prefect server). This change allows for the UI
to be customized by providing an extra env variable named
`VUE_APP_REQUEST_CREDENTIALS_MODE`, which will cause the UI to forward cookies
to the Apollo server if the variable is set with `include` as its value. If the
variable is left unset, the behaviour is the same as before.

### DIFF
--- a/tests/unit/vue-apollo.spec.js
+++ b/tests/unit/vue-apollo.spec.js
@@ -1,0 +1,21 @@
+describe('VueApollo', () => {
+  beforeEach(() => {
+    jest.resetModules()
+  })
+
+  it('does not include credentials by default', () => {
+    const { defaultOptions } = require('@/vue-apollo')
+    const { link } = defaultOptions
+
+    expect(link.credentials).toEqual(undefined)
+  })
+
+  it('includes credentials when specifically indicated', () => {
+    process.env.VUE_APP_REQUEST_CREDENTIALS_MODE = 'include'
+
+    const { defaultOptions } = require('@/vue-apollo')
+    const { link } = defaultOptions
+
+    expect(link.credentials).toEqual(undefined)
+  })
+})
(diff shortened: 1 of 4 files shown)


tests/unit/vue-apollo.spec.js
```diff
@@ -0,0 +1,21 @@
...
+    expect(link.credentials).toEqual(undefined)
+  })
+
+  it('includes credentials when specifically indicated', () => {
+    process.env.VUE_APP_REQUEST_CREDENTIALS_MODE = 'include'
+
+    const { defaultOptions } = require('@/vue-apollo')
+    const { link } = defaultOptions
+
+    expect(link.credentials).toEqual(undefined)
```
stackoverfloweth, line 19: this assertion seems incorrect?

stackoverfloweth (commented):
Looks good to me 👍

tests/unit/vue-apollo.spec.js
jverce: It is, and I'm working on a proper test case at the moment. I'll let you know once that's done. Thanks!

sjmccallen:
I'm interested in this feature as well. We would like to authenticate users to the UI and API using a proxy hooked up to our authentication. Does this feature mean that the access cookie/token granted by authenticating to the UI end point would be forwarded to the API endpoint when using the UI through the browser?

My understanding is a bit limited right now, but I hope to understand more. 😃 

Thanks!

jverce:
You're right @sjmccallen, being able to forward the authenticaton cookie to both the Prefect UI server and the Apollo server is what would allow you to have the authentication in place.

However, we found a somewhat cleaner way of doing this without touching the UI source code. We basically have the same domain for both the UI and the Apollo service (i.e. the GraphQL API), with the following proxy/routing rules:
- Requests going to `prefect.mydomain.com/graphql` are forwarded to the Apollo service
- Any other request to `prefect.mydomain.com` is forwarded to the original UI server

That way, the browser will always forward the cookies (since everything points to the same domain) and we can have these 2 separate services being properly used.

Issue #1113, opened 2021-11-05
Improve Agent Automations UX 


joeclayallday:
**Why**
As a user, when I make Agent Automations the installation instructions could be shown more obviously and I'd like to not leave the automations page to copy the agent-config-id (currently I have to leave and ping graphQL for agent-config-ids).

**What**
• After creating an agent config, allow users to simply copy the newly created agent-config-ids in the UI.
• Add clear text instructions that inform users how to install these agent automations at registration in the Prefect CLI & that if agents share agent-config-ids, then the automations doesn't work until all agents go unhealthy instead of individually.
• This proposal doesn't change any application behavior, only front-end tweaks.

**Notes**
Proposed in-app messaging:

To set up an agent automation, you need to add the agent-config-id flag at agent registration.
If you attach the same ID to multiple agents, Prefect will only notify you if all agents are unhealthy.

**BEFORE**
<img width="861" alt="Screen Shot 2021-11-05 at 5 10 47 PM" src="https://user-images.githubusercontent.com/4295811/140584126-f76966f9-8f17-4ddf-9497-5fa7fe286de1.png">

**AFTER**
<img width="859" alt="Screen Shot 2021-11-05 at 5 10 50 PM" src="https://user-images.githubusercontent.com/4295811/140584139-013eb803-371f-4f88-b6a2-49adcc95037c.png">

**NOTE:** On Create Success Modal, reveal agent-config-id

<img width="1125" alt="Screen Shot 2021-11-05 at 7 55 26 PM" src="https://user-images.githubusercontent.com/4295811/140592494-bfbeee9e-7010-4e72-84b8-60c69f2e330b.png">

<img width="857" alt="Screen Shot 2021-11-05 at 5 10 55 PM" src="https://user-images.githubusercontent.com/4295811/140584148-eeb9e1e2-6c95-4e59-a129-5dfe496f5c93.png">

<img width="824" alt="Screen Shot 2021-11-05 at 5 10 59 PM" src="https://user-images.githubusercontent.com/4295811/140584154-474df2d2-ab69-4bac-88d6-2444e06bd4da.png">

<img width="816" alt="Screen Shot 2021-11-05 at 5 11 02 PM" src="https://user-images.githubusercontent.com/4295811/140584161-e5fe0de9-1b59-4c80-b90e-9c5a239b903d.png">

zhen0:
[...]

PR #1153, opened 2021-12-09
Feature: Enable deletion of agent configs in automations

zhen0:
## Describe this PR
Begins to add suggestions from #1113


ThatGalNatalie:
After creating an agent config and then deleting it, the next button was still clickable for me and the config name was still highlighted. I expected the next button to be disabled and the config name to be changed back to "this config"

https://user-images.githubusercontent.com/40726391/145506459-96017d17-5e4b-4ae1-8617-2f90073730f8.mov

stackoverfloweth:
`pages/Dashboard/Automations/AddAutomationCard.vue` is a huge file. I think it would be good to move your changes out into a separate file. If it's not too out of scope, I'm sure there are other components that could be picked out without a huge refactor

stackoverfloweth:
very nice 👌

stackoverfloweth (approved):
very nice 👌

cicdw (approved):

PR #1158, opened 2021-12-14
rolling back watcher change

stackoverfloweth:
PR Checklist:

### DIFF
--- a/src/layouts/ManagementLayout.vue
+++ b/src/layouts/ManagementLayout.vue
@@ -16,14 +16,14 @@ export default {
       required: false
     }
   },
-  computed: {
-    showPage: {
-      get() {
-        return this.show
-      },
-      set(value) {
-        this.$emit('update:show', value)
-      }
+  data() {
+    return {
+      showPage: this.show || false
+    }
+  },
+  watch: {
+    show(val) {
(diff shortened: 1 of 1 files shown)


src/layouts/ManagementLayout.vue
```diff
@@ -16,14 +16,14 @@ export default {
       required: false
     }
   },
-  computed: {
-    showPage: {
-      get() {
-        return this.show
-      },
-      set(value) {
-        this.$emit('update:show', value)
```
znicholasbrown, line 25: Are any higher-level components dependent on this emitter?

znicholasbrown: Resolved live

znicholasbrown (approved):

src/layouts/ManagementLayout.vue
stackoverfloweth: this is a rolling back to previous functionality, so no, nothing should be expecting to receive this emit. In an attempt to remove the watcher, I mistakingly assumed that the prop was always provided, so updating the prop would be a more correct way to ensure that `show` stays up to date. Views like the FlowConcurrency do not pass the prop, so emitting update:show has no effect

PR #1166, opened 2021-12-14
BugFix: removed unused onIntersect methods

stackoverfloweth:
We removed several v-intersect directives without cleaning up the callback methods on the components, this PR just cleans those up


zhen0 (approved):
Thanks @stackoverfloweth

PR #1173, opened 2021-12-15
added http-security-headers to apollo graphql network traffic

stackoverfloweth:
should resolve https://github.com/PrefectHQ/devsecops/issues/5


stackoverfloweth:
@cicdw @dwildes-prefect this PR adds the missing security http headers to all graphql network requests

Screen Shot 2021-12-14 at 10 07 09 PM

Do either of you have any concerns with the values I'm using? I picked the most restrictive in the short amount of research I've done. Perhaps @znicholasbrown would have opinions as well? 

`x-frame-options: deny` - will prevent the page from rendering <frame>, <iframe>, <embed> or <object>
`x-content-type-options: nosniff` - no idea what negative effects this has, or why anybody wouldn't want to use nosniff
`x-permitted-cross-origin-domain-policies: none` - seems to work like Access-Control-Allow-Origin except it's only for Adobe flash or pdf content
`content-security-policy: default-src 'self'` - apparently controls what resources the user agent is allowed to load for the page, default-src 'self' is the most restrictive, all content has to come from the site's own origin (this excludes subdomains.) 

Locally, with these headers on all my graphql queries things seem to load fine, including logging in. Highly recommend more testing in more environments.

cicdw:
I don't see any issue with those values -- the only one I could see potentially causing an unexpected issue is the content security policy but seems like it's working fine

zhen0 (commented):
This looks fine to me but I would welcome reviews from @dwildes-prefect and @znicholasbrown before we merge.

znicholasbrown (approved):

cicdw (approved):

PR #1181, opened 2021-12-22
BugFix: Improve error handling on restart button and refetch state after restart

### LINKED ISSUE #1067: Prevent Navigation / Provide Positive Alert when a Flow Run has been successfully restarted
**Why**

[Session](https://app.logrocket.com/qetfiw/cloud-web-ui/s/5-eeb5451d-dae5-40ef-aa5c-4ca23a0b7ce0/0?filters=%255B%257B%2522id%2522%253A%2522HJRSqnzNK%2522%252C%2522type%2522%253A%2522email%2522%252C%2522operator%2522%253A%257B%2522name%2522%253A%2522is%2522%252C%2522type%2522%253A%2522IS%2522%252C%2522hasStrings%2522%253Atrue%252C%2522autocompleteEnabled%2522%253Atrue%257D%252C%2522strings%2522%253A%255B%2522dylan%2540prefect.io%2522%255D%257D%255D&fromTab=latest&t=1632975428968)
For Prefect Cloud, when I press the `Restart` button, I am often unsure if my Flow Run is actually restarting or not. I generally end up pushing the button more than once. When I push the button and I'm unsure if the run has restarted, the UI lets me navigate to other pages. I would be happy for it to stop me (like gmail does when I have unsaved changes to my inbox) so that I'm certain that the flow run is restarting like I expect.
**What**

* [ ] The user arrives at a flow run page where the flow run has failed
* [ ] The user presses restart
* [ ] The user attempts to navigate away before the restart is fully processed, and is prevented

**Notes**

Total UX improvement, not a bug

### LINKED ISSUE #1085: Restarting Quick Run from Failed state doesn’t always restart the FlowRun
## Opened from the [Prefect Public Slack Community](https://prefect.io/slack)

**tony.yun**: btw, I’ve seen this a while that I have to restart the flow twice to make it really restarted. The first time click just logged a line but nothing in effect until the second time click. Is this a known issue?

**tony.yun**: 

**anna**: Hi <@U029S51BLA2>, thanks for reporting that. I was able to reproduce the issue and indeed, something seems weird. Here is an example flow I used.
First setting a key in KV Store:

```
from prefect.backend import set_key_value
key_value_uuid = set_key_value(key="fail_or_not", value="fail")
```

Then failing the flow based on the value from KV store:

```
from prefect.backend import get_key_value

@task(log_stdout=True)
def fail_if_key_say_so():
    value = get_key_value(key="fail_or_not")
    if value == "fail":
        raise ValueError("Failing the task...")
    print("Success!")

with Flow("fail_based_on_key") as flow:
    fail_if_key_say_so()
```

Changing the value should result in success upon restart:

```
from prefect.backend import set_key_value
key_value_uuid = set_key_value(key="fail_or_not", value="not")
```

But similarly to you, I had to click twice for the restart to take effect. *There is a caveat though*. This happens ONLY when the original flow is triggered through “Quick Run” in the UI. When I trigger the flow run from CLI or when the flow runs on schedule, the Restart button in the UI triggers a restart immediately.
<@ULVA73B9P> open “Restarting Quick Run from Failed state doesn’t restart the FlowRun until the user clicks the Restart button twice” in UI

**anna**: <@ULVA73B9P> open “Restarting Quick Run from Failed state doesn’t always restart the FlowRun”

**kevin701**: <@ULVA73B9P> open “Restarting Quick Run from Failed state doesn’t always restart the FlowRun” in UI

Original thread can be found [here](https://prefect-community.slack.com/archives/CL09KU1K7/p1634236568382300?thread_ts=1634236568.382300&cid=CL09KU1K7).

### LINKED ISSUE #1147: Refetch flow run state after restart
**Why**

Currently when a flow run is restarted, its state does not update and it's possible to restart the flow a second time.  

**What**

* [ ] When the restart dialog closes the flow run state updates
* [ ] 
**Notes**

zhen0:
PR Checklist:

## Describe this PR
Updates the error handling for the flow run restart button and ensures that utility downstream tasks are loaded before the restart can be confirmed. Also calls update once the set flow run state mutation has completed. 

Closes #1147 Closes #1085 Closes #1067


src/pages/FlowRun/Restart-Dialog.vue
```diff
@@ -106,32 +112,34 @@ export default {
...
-
+            this.$emit('update')
             this.setAlert({
               alertShow: true,
               alertMessage: 'Flow run restarted.',
               alertType: 'success'
             })
           } else {
             this.error = true
+            this.errorMessage = 'There was an error with the API call'
```
zhen0, line 123: I'm not a fan of this set of if/else and the error messages but I'm not sure I can handle it better without a bigger refactor.  I've added in more specific error messages to help users (and us) with any future debugging.

stackoverfloweth (approved):

PR #1183, opened 2021-12-22
Bugfix: Add log rocket prop

zhen0:
@stackoverfloweth and @cicdw as discussed


src/pages/TeamSettings/Secrets.vue
```diff
@@ -385,6 +385,7 @@ export default {
       :title="isSecretUpdate ? 'Modify Secret' : 'Create New Secret'"
       @cancel="resetSelectedSecret"
       @confirm="validateAndSetSecret"
+      data-private
```
cicdw, line 388: Can we put this on `secretValueInput` below as well?

cicdw (approved):

PR #1187, opened 2021-12-23
BugFix: Add once to intersect on lastTenRuns

### LINKED ISSUE #1179: run history should not reload on scroll
**Why**
Parts of the the UI can be on the slow side; loading run history is one of them. When scrolling through a list of flows, the run history should not have to reload as the user scrolls around.

**What**
* once the run history is loaded for a flow - leave it loaded - don't reload it after it goes offscreen and then back on screen.

**Notes**
Untitled

zhen0:
Only call onIntersect on mount and once intersected to enable easier scrolling through the flow table.  This potentially adds more data and polling but should still remove the big initial poll of pulling everything on the page even when not visible.  Based on vuetify docs here: https://vuetifyjs.com/en/api/v-intersect/#argument

Closes #1179

### DIFF
--- a/src/components/LastTenRuns.vue
+++ b/src/components/LastTenRuns.vue
@@ -177,7 +177,7 @@ export default {
 <template>
   <div>
     <BarChart
-      v-intersect="{ handler: onIntersect }"
+      v-intersect.once="onIntersect"
       :items="preppedFlowRuns"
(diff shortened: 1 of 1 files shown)


stackoverfloweth (approved):
Great solution

PR #1188, opened 2021-12-23
BugFix: Ensure label warning updates 

### LINKED ISSUE #1182: The flow run page shows an issue with label even though the agent is healthy (latency issue)
The flow run page shows an issue with a label for a short while, but as soon as we hover over the red icon, it disappears. 

ui_issue_2

zhen0:
Refetch flow run on label edit and start polling on label warning intersect not input because we need polling to see if warning is required

Closes #1182

### DIFF
--- a/src/pages/FlowRun/Details-Tile.vue
+++ b/src/pages/FlowRun/Details-Tile.vue
@@ -79,6 +79,9 @@ export default {
         space_in_empty_paren: true,
         preserve_newlines: false
       })
+    },
+    handleRefetch() {
+      this.$emit('refetch')
     }
   }
 }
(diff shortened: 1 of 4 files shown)


src/pages/FlowRun/Details-Tile.vue
```diff
@@ -79,6 +79,9 @@ export default {
         space_in_empty_paren: true,
         preserve_newlines: false
       })
+    },
+    handleRefetch() {
+      this.$emit('refetch')
```
stackoverfloweth, line 84: I think it would be better if this component didn't listen for an event only to emit bubble it up to it's parent. It's probably more of a refactor than is responsible, but ideally we would probably extract the apollo mutations out of LabelEdit so the parent was responsible for actually executing the query, LabelEdit could be a dumb component like a select-box. It could just use a v-model, and when that model changes the parent can trigger the mutation and emit up to it's parent 1 layer that a refetch is advised. 

Looking at LabelEdit, it appears to have some complexity trying to be the LabelEditor for both flows and tasks, which would also be cleaned up by this refactor. But as I mentioned before, this is probably not a responsible change at the moment.

stackoverfloweth (approved):

PR #1197, opened 2022-01-03
Bugfix: Add loading state to the cloud hooks table 

### LINKED ISSUE #1167: BUG: Cloud hooks page loading state
**Bug Description**

The cloud hooks page in the team menu, currently has no loading state.  The page briefly shows "no cloud hooks found" before actually showing cloud hooks. 

**Steps To Reproduce**

1. From the dashboard click on team > cloud hooks
2. Observe the "no cloud hooks found" message flicker up and then be replaced by a list of cloud hooks.  You can also refresh the page to see again. 

**Browsers Tested:**
* [x] Chrome
* [ ] Firefox
* [ ] Safari
* [ ] Edge
* [ ] IE

----------
**Notes**

https://user-images.githubusercontent.com/40272060/147179432-74f157a9-b8c8-4468-a80d-4144d79ec694.mov

ThatGalNatalie:
## Description
<! -- What is it meant to do? -->
Adds a loading state to the cloud hooks table 

## Linked Issues
<! -- Use a key word (e.g. closes or resolves) to close related issues  -->
Resolves #1167  

## Tests and performance

 - [x] Changes to any .js files are covered by existing tests or this PR adds new tests (if not please explain why below)
 - [x] No new packages are added or package size and performance considerations are discussed below
 - [x] PR Title fits our [changelog format](https://github.com/PrefectHQ/ui#readme)

 ## Releases/Changelog cuts only
 - [ ] Completed the [QA Checklist](https://www.notion.so/prefect/Cloud-UI-QA-Checklist-038a5a5d40a249d78232917ad1b923b9) in staging

### DIFF
--- a/src/pages/TeamSettings/CloudHooks.vue
+++ b/src/pages/TeamSettings/CloudHooks.vue
@@ -354,6 +354,7 @@ export default {
           prevIcon: 'keyboard_arrow_left',
           nextIcon: 'keyboard_arrow_right'
         }"
+        :loading="!cloudHooks"
         no-data-text="No Cloud Hooks found."
         ><!-- HEADERS -->
         <template #header.name="{ header }">
--- a/src/pages/TeamSettings/CloudHooks.vue
+++ b/src/pages/TeamSettings/CloudHooks.vue
@@ -73,7 +73,8 @@ export default {
           sortable: false,
           width: '20%'
         }
-      ]
+      ],
+      hooksLoaded: false
     }
   },
   computed: {
@@ -255,7 +256,7 @@ export default {
 </script>
 
 <template>
-  <ManagementLayout :show="!isLoadingTable" control-show>
+  <ManagementLayout show control-show>
     <template #title>Cloud Hooks</template>
     <template #subtitle>
       <div v-if="!hasPermission('update', 'cloud-hook')">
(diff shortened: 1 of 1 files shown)


src/pages/TeamSettings/CloudHooks.vue
```diff
@@ -354,6 +354,7 @@ export default {
           prevIcon: 'keyboard_arrow_left',
           nextIcon: 'keyboard_arrow_right'
         }"
+        :loading="!cloudHooks"
```
stackoverfloweth, line 357: I'm not sure the lack of `cloudHooks` means the component is "loading". I would use the more typical approach of defining a `loadingKey` on the apollo query (line 227) and defining loading as `loadingKey > 0`

stackoverfloweth (changes requested):

src/pages/TeamSettings/CloudHooks.vue
ThatGalNatalie: @stackoverfloweth I added the `loadingKey` option to the query but it's not updating the data property that I passed it. When I reload the page, the property stays at 0 but I'm not sure why? I found that we already had a `isLoadingTable` computed property and used that instead which sets the return value to `true` in the result hook. Is that okay too?

stackoverfloweth: Hmm yeah, in my testing I had the same result. That's really weird that this query doesn't seem to work well with the standard `loadingKey` like the others. The thing I don't like about using the `isLoadingTable` is that it too just becomes false only when there are results, so if there were no items I believe it would appear to load forever.  But after spending a good hour or so with no better solution I'm fine with just moving forward as is

stackoverfloweth (approved):

PR #1250, opened 2022-03-09
Feature: Adds Salesforce Chat Bot

space-age-pete:
## Description
Adds the Salesforce Chat Bot, which provides either live chat, or recommends articles, based on the user's Salesforce Account ID, which is included in the "terms" field in the License table. This code comes from a modified code snippet provided by Salesforce. The Content-Security-Policy meta tag in the index.html file is updated to allow this embed, and some css is also added to the same file.

## Tests and performance

 - [X] Changes to any .js files are covered by existing tests or this PR adds new tests (if not please explain why below)
 - no changes to .js files
 - [X] No new packages are added or package size and performance considerations are discussed below
 - [X] PR Title fits our [changelog format](https://github.com/PrefectHQ/ui#readme)


src/App.vue
```diff
@@ -366,6 +367,94 @@ export default {
         return data.pendingInvitations
       }
     })
+
```
znicholasbrown, line 370: This is the wrong location for this logic - it should most likely be in its own file in the `plugins` directory and initialized in `main.js` after checking that we're in a cloud environment (the node_env shouldn't matter)

src/App.vue
```diff
@@ -366,6 +367,94 @@ export default {
...
+            team: '',
+            url1: '',
+            url2: '',
+            baseLiveAgentContentURL: '',
+            deploymentId: '',
+            buttonId: '',
+            baseLiveAgentURL: ''
+          }
+
+    var initESW = function(gslbBaseURL) {
```
znicholasbrown, line 422:
```suggestion
    const initESW = function(gslbBaseURL) {
```

src/App.vue
```diff
@@ -366,6 +367,94 @@ export default {
...
+        deploymentId: deploymentId,
+        buttonId: buttonId,
+        baseLiveAgentURL: baseLiveAgentURL,
+        eswLiveAgentDevName: team,
+        isOfflineSupportEnabled: false
+      })
+    }
+
+    if (!window.embedded_svc) {
+      var s = document.createElement('script')
```
znicholasbrown, line 446:
```suggestion
      const s = document.createElement('script')
```

src/App.vue
```diff
@@ -366,6 +367,94 @@ export default {
...
+            sfcid: '00D2E000000o3Ut',
+            team: 'Chat_Service_Team',
+            url1: 'https://prefect.my.salesforce.com',
+            url2: 'https://prefect.force.com/partners',
+            baseLiveAgentContentURL:
+              'https://c.la1-c2-ia2.salesforceliveagent.com/content',
+            deploymentId: '5722E0000000dd5',
+            buttonId: '5732E0000000eF1',
+            baseLiveAgentURL:
+              'https://d.la1-c2-ia2.salesforceliveagent.com/chat'
```
znicholasbrown, line 409: Maybe these should pull from environment variables and only live in CICD?

public/index.html
```diff
@@ -13,11 +13,11 @@
...
+      script-src-elem 'self' 'unsafe-eval' 'unsafe-inline' https://cdn.lr-ingest.io https://js.stripe.com https://netlify-cdp-loader.netlify.app/netlify.js https://prefect.my.salesforce.com/ https://prefect.force.com/ https://prefect.my.salesforce.com https://c.la1-c2-ia2.salesforceliveagent.com/ https://d.la1-c2-ia2.salesforceliveagent.com/ https://static.lightning.force.com https://prefect--partial.my.salesforce.com https://c.la4-c1cs-phx.salesforceliveagent.com/ https://d.la4-c1cs-phx.salesforceliveagent.com/ https://partial-prefect.cs90.force.com/ https://d.la3-c1cs-ia2.salesforceliveagent.com/ https://service.force.com/ https://prefect--aleysian.my.salesforce.com https://aleysian-prefect.cs29.force.com/;
       style-src 'self' 'unsafe-inline';
       font-src data: 'self';
-      frame-src https://*.prefect.io https://prefect.auth0.com https://js.stripe.com https://cloud.prefect.io https://*.okta.com;
+      frame-src 'self' * data: https://*.prefect.io https://prefect.auth0.com https://js.stripe.com https://cloud.prefect.io https://*.okta.com;
```
znicholasbrown, line 20: Is a wildcard correct here?

znicholasbrown (changes requested):

.circleci/config.yml
```diff
@@ -193,6 +209,12 @@ jobs:
     docker:
       - image: node:14.7
     steps:
+      - run:
+          name: Load salesforce_env file
+          command: |
+            cat .salesforce_env_prod | while read line; do
+              export $line
```
jawnsy, line 216: GitHub has a magic file you can write to, in order to set env vars (`$GITHUB_ENV`), so you might be able to just cat into that instead: https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-environment-variable

jawnsy: ah nevermind, this is for Circle CI

znicholasbrown (approved):

PR #1324, opened 2022-08-31
Revert "Bump terser from 4.8.0 to 4.8.1"

zhen0:
Reverts PrefectHQ/ui#1308


zhen0:
IAPI also broken here

PR #1338, opened 2022-09-20
Feature: adapts chatbot to match new SFDC changes

space-age-pete:
## Description
The salesforceChat plugin now sends the User ID to the chatbot rather than the Salesforce Account ID, which is now required after the backend of the chat bot changed in SFDC. 

New URLs were added to the index.html file in order to allowlist the SFDC URLs associated with the version of the chatbot in the SFDC Partial Sandbox environment, which is used for testing.

## Tests and performance

 - [x] Changes to any .js files are covered by existing tests or this PR adds new tests (if not please explain why below)
 - [x] No new packages are added or package size and performance considerations are discussed below
 - [x] PR Title fits our [changelog format](https://github.com/PrefectHQ/ui#readme)


stackoverfloweth (approved):

dylanbhughes:
Hey @zhen0 and @znicholasbrown just a heads up re: timing for related changes -- I think we're going to move to a place in the Looker where we're going to only expose masked or redacted PII to everyone. This means that we'll need to push changes in SFDC that are coupled with this PR to prod in order for folks to be able match Contacts/Leads -> Cloud2 objects via Cloud2 IDs. We have to turn on the masking by EOD Friday for SOC2 audit reasons. If these changes don't make it to SFDC prod by then, we will have a period when the GTM org will be unable to relate Contacts/Leads to Cloud2 information in some cases, resulting in poor reporting. 

If you have the bandwidth to assess these changes before the next Cloud 2 production release, we'd be most appreciative 🙏

znicholasbrown (approved):

PR #1365, opened 2023-03-07
Bugfix: Include STANDARD licenses in EOL support banner

znicholasbrown:
## Description
Adds a missing plan


znicholasbrown:
Yeah that's right @zhen0, we'll deploy this manually

PR #1392, opened 2024-01-09
Chore: Add productionSourceMap setting

znicholasbrown:
## Description
#1391 removed the webpack-level source map setting but I didn't realize there's a vue-cli-service level setting for this as well. Hopefully this'll address that.


znicholasbrown:
Ok verified in the preview build that this does what we're hoping

zhen0 (approved):

